PR #36, opened 2022-04-29
Stream zip file immediately

hillmich:
Instead of generating the whole zip file first, start streaming immediately.


hillmich:
The new commit make the generator free the memory after reading the chunks for sending.

nquetschlich (approved):

PR #37, opened 2022-04-29
Remove in memory zipping

nquetschlich:
With this PR, the zipping is done again by saving the generated zip files on storage instead of in memory only to avoid server problems if several users download all files at the same time.


burgholzer (approved):
LGTM

PR #46, opened 2022-05-04
Server sync adaptions

nquetschlich:
- removed hint for timeout
- added the config.json to the rsync because it is needed for #num benchmarks calculation for algo level


burgholzer (approved):
LGTM

PR #51, opened 2022-05-30
Logging

nquetschlich:
Added simple logging mechanism for

- pre-generated zip with all files
- custom selection of benchmark including the selection parameters and number of downloaded benchmark files


burgholzer (approved):
LGTM.
In the long run we might want to have some sort of download statistic, but this is something for another PR 👍🏼

Issue #52, opened 2022-05-31
Copy the ZIP file first


hillmich:
In the deployment workflow, the sources are copied first, followed by the the zip file.
https://github.com/cda-tum/MQTBench/blob/0928aeb78ba2e9f90c5e007f9a2734133c14a3d2/.github/workflows/server_deployment.yml#L44-L49

This triggers a reload on the server before the new zip is available, thus the zip should be copied first.

PR #60, opened 2022-06-07
👷changed the sequence order of the rsync commands to trigger the serv…

nquetschlich:
changed the sequence order of the rsync commands to trigger the server restart after all necessary files have been synced

--> See Issue #52


hillmich (approved):
LGTM

Issue #55, opened 2022-06-01
Make Benchmark Generation Reproducible


hillmich:
The generation of benchmarks is currently not reproducible. Two changes should enable reproducible generation

1. [x] Set a seed for the transpiler (Qiskit supports this, I don't know about tket)
2. [x] Fix the versions in Python dependencies (explicitly use `==` instead of `~=`)

Bonus: Check the actually required packages for bench and benchviewer (and delete not required packages).

nquetschlich:
solved in #93

Issue #49, opened 2022-05-16
Add optimization_level for get_one_benchmark and independent layer


nquetschlich:
To be consistent, optimization_level=1 needs to be added to get_one_benchmark method for the independent layer.

- [ ] Parameter added
- [ ] Created new release afterwards

PR #62, opened 2022-06-07
Transpile calls

nquetschlich:
Pull request for:
Issues #55 and #49

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,13 +27,12 @@ setup(
     url="https://github.com/cda-tum/mqtbench",
     keywords="mqt quantum benchmarking performance testing",
     install_requires=[
-        "matplotlib==3.5.1",
         "qiskit==0.35",
+        "qiskit[visualization]==0.35.0",
         "pandas==1.3.5",
         "flask==2.1.2",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -27,13 +27,12 @@
...
     install_requires=[
-        "matplotlib~=3.5.1",
-        "qiskit~=0.35",
-        "pandas>=1.3.5,<1.5.0",
-        "flask>=2.0.2,<2.2.0",
-        "networkx>=2.7.1,<2.9.0",
-        "pytest~=7.1.1",
-        "pylatexenc~=2.10",
+        "qiskit==0.35",
+        "qiskit[visualization]==0.35.0",
```
hillmich, line 31: `qiskit[visualization]==0.35.0` should be be enough (no need to include `qiskit==0.35.0` as well).

hillmich (changes requested):

setup.py
nquetschlich: Thanks for that feedback, I commited the change.

hillmich (approved):
LGTM

Issue #82, opened 2022-08-18
Remove "Excited State" Benchmark


nquetschlich:
Benchmark should be removed since deprecated methods are used.

nquetschlich:
solved in #93

PR #84, opened 2022-08-19
Remove excitedstate

nquetschlich:
solves #82


burgholzer (changes requested):
Why are there so many commits in here that are not related to the PR at hand?
There seems to be something wrong with the commit the branch was based on. 
Please consider fixing this, so that the  PR only contains the dedicated changes.

The last commit itself LGTM. So the PR would be ready to merge.

Issue #86, opened 2022-08-24
Provide Interfaces for MQT Predictor


nquetschlich:
The MQT Bench compilation pipelines should be made available such that 

- [ ] provided with a `qasm` file and a target location, the chosen mapped version is created.

nquetschlich:
solved in #93

PR #91, opened 2022-08-29
Provide compilation interface

nquetschlich:
PR for solving issue #86


config.json
```diff
@@ -1,5 +1,5 @@
 {
-"timeout": 100,
+"timeout": 1,
```
burgholzer, line 2: is this timeout change intentional?

mqt/bench/tests/test_all.py
```diff
@@ -604,3 +599,39 @@ def test_configure_end():
...
+def test_saving_qasm_to_alternative_location_with_alternative_filename(
+    abstraction_level: int,
+):
+    directory = "."
+    filename = "ae_test_qiskit"
+    qc = get_one_benchmark("ae", abstraction_level, 5)
+    assert qc
+    res = qiskit_helper.get_mapped_level(
+        qc, "ibm", qc.num_qubits, "ibm_washington", 1, False, False, directory, filename
+    )
```
burgholzer, line 621: just a hint for the future (does not necessarily have to be adopted right here and now):
for function calls like these it would really help to use keyword arguments, so that the meaning of each of the parameters is more clear. I can't really tell (from a reviewer perspective) whether the arguments supplied to these functions "make sense".

mqt/bench/tests/test_all.py
```diff
@@ -604,3 +599,39 @@ def test_configure_end():
...
+    assert res
+    path = os.path.join(directory, filename) + ".qasm"
+    assert os.path.isfile(path)
+    os.remove(path)
```
burgholzer, line 625: Also nothing necessary to change right here and now, but I believe `os.path` is more and more being replaced by usage of the `pathlib` module (https://docs.python.org/3/library/pathlib.html). You might want to look into that for the future.

mqt/bench/tests/test_all.py
```diff
@@ -604,3 +599,39 @@ def test_configure_end():
...
+
+    directory = "."
+    filename = "ae_test_tket"
+    qc = get_one_benchmark("ae", abstraction_level, 7)
+    assert qc
+    res = qiskit_helper.get_mapped_level(
+        qc, "ibm", qc.num_qubits, "ibm_washington", 1, False, False, directory, filename
+    )
```
burgholzer, line 633: the filename here says `tket` but the function uses the `qiskit_helper`. This does not seem right.

burgholzer (changes requested):
A couple of minor comments but mostly this LGTM 👍🏼

config.json
nquetschlich: In this case, it was more or less intentional. So far, the creation of all benchmarks has not been executed and I played around with the timeout limit for testing purposes. As a consequence, the actual number is not relevant right now and both `100` and `1` are equally "correct".

mqt/bench/tests/test_all.py
nquetschlich: Thanks for that feedback. I will do that in the future.

mqt/bench/tests/test_all.py
nquetschlich: I will check it out and consider it in the future.

mqt/bench/tests/test_all.py
nquetschlich: Thanks for pointing that out, it is fixed now.

config.json
burgholzer: Sounds reasonable to me 👍🏼

burgholzer (approved):
LGTM 👍🏼

PR #107, opened 2022-09-06
📦 Update Dependencies

burgholzer:
This PR updates all the dependencies in MQT Bench and how dependencies themselves are organized.
Some of the dependencies were pinned to specific version before but the package functionality itself was not influenced by the particular version.
As a consequence, these dependency versions have been loosened to supported ranges.


burgholzer:
Two of the relevant PRs have been merged already. Just waiting on new versions of the packages to be released.
These alone, together with the workaround in this PR should allow to use the most recent Qiskit packages.
If https://github.com/CQCL/tket/issues/506 were to be implemented, this would also remove the need for patching the OQC QASM files afterwards (at least for circuits generated with tket)

burgholzer:
At some point we really should try to get this merged. Something for after the  deadline.

nquetschlich:
I think we got all things from this PR covered in #144. 
@burgholzer: I you agree, please close this PR.

Issue #109, opened 2022-09-10
Filtering the Database is very slow (especially on the web interface)


nquetschlich:
Currently, the filtering logic and, as a consequence, the time it takes to update the number of currently selected benchmarks takes rather long to refresh.  One reasons for that lies within the MQT Bench UI solution of v0.1.0rc1, where the selected levels and corresponding options were selectable for each and every benchmark while in v0.1.0 this feature was omitted to foster simplicity and usability. 

However, the filtering logic `filterDatabase(...)` in `benchviewer/src/backend.py` still iterates over each benchmark and filters accordingly. This logic should be adjusted such that first the whole database of all available benchmarks should be filtered for the selected qubit range followed by all selected levels and corresponding options.

PR #110, opened 2022-09-10
✨ improved filter logic for speed-up

nquetschlich:
solves #109


burgholzer (approved):
Looks like a small, but rather effective change 👍🏼 
If I were to suggest something for the moment, it would be to pick better variable names than `db_tmp[123456]`. However, feel free to merge this as it is 👍🏼 

In the long run, we might need a more generic version of this filtering procedure to avoid adding an endless amount of loops and ifs when new options/compilers are added.

Issue #95, opened 2022-09-01
Add postprocessing for all OCQ benchmarks


nquetschlich:
Right now, the qasm export is faulty for OQC files. The postprocessing must be implemented for the `get_one_benchmark`method as well.

PR #138, opened 2022-11-28
Fix oqc qasm

nquetschlich:
It was meant to solve #95 but actually the problem was never present for the `get_one_benchmark` method since that method always returns the Qiskit::QuantumCircuit object. Nevertheless, the OQC error handling is improved:

- It is done now immediately after creation.
- The `opaque` function of OpenQASM is now used to avoid introducing non-native gates in the native and mapped levels.

### DIFF
--- a/mqt/bench/utils/utils.py
+++ b/mqt/bench/utils/utils.py
@@ -6,7 +6,6 @@ from datetime import date
 
 import networkx as nx
 import numpy as np
-from joblib import Parallel, delayed
 from pytket import __version__ as __tket_version__
 from qiskit import QuantumCircuit, __qiskit_version__
 from qiskit.algorithms import EstimationProblem
@@ -249,6 +248,9 @@ def save_as_qasm(
         f.write("\n")
         f.write(qc_str)
     f.close()
+
+    if "oqc" in filename:
+        postprocess_single_oqc_file(filename)
     return True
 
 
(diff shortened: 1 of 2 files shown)


mqt/bench/utils/utils.py
```diff
@@ -308,55 +310,40 @@ def get_molecule(benchmark_instance_name: str):
...
-        delayed(postprocess_single_oqc_file)(directory, filename)
-        for filename in os.listdir(directory)
-    )
-
+# def postprocess_oqc_qasm_files():
+#     directory = get_qasm_output_path()
+#     Parallel(n_jobs=-1, verbose=100)(
+#         delayed(postprocess_single_oqc_file)(directory, filename)
+#         for filename in os.listdir(directory)
+#     )
```
burgholzer, line 318: if this is no longer needed, then it might as well be removed.

mqt/bench/utils/utils.py
```diff
@@ -308,55 +310,40 @@ def get_molecule(benchmark_instance_name: str):
...
 
-def postprocess_single_oqc_file(directory: str, filename: str):
 
-    f = os.path.join(directory, filename)
-    if "oqc_lucy_qiskit" in f or "oqc_qiskit" in f:
-        with open(f) as f:
+def postprocess_single_oqc_file(filename: str):
+    if "mapped_oqc_lucy_qiskit" in filename or "nativegates_oqc_qiskit" in filename:
```
burgholzer, line 322: This is very specific to the OQC Lucy device and, hence, might require further adaptation in the future if other devices are added. Wouldn't it make more sense to just check whether the filename contains `oqc` and `mapped` or `nativegates`? And then make a distinction between `qiskit` and `tket`. This might also help to eliminate some duplicate code between both of the branches.

burgholzer (approved):
LGTM 👍🏼  Just a small suggestion on how to improve the code for the post-processing a bit.

burgholzer:
This still fixes #95, right?

nquetschlich:
> This still fixes #95, right? If so, then I'd link the issue to this PR so that it gets closed automatically once this PR is in.

Actually no, because #95 was never a problem, since the ECR problem happens only when the qasm files are saved. The `get_one_benchmark` method is not saving the qasm files but just returning the Qiskit/TKET Circuit object. But nevertheless, this PR improves the postprocessing since it is now done automatically after the file is created. 

What do you mean by "linking"? Can you do it explicitly or is it done by just referencing somewhere to the issue?

burgholzer:
> > This still fixes #95, right? If so, then I'd link the issue to this PR so that it gets closed automatically once this PR is in.
> 
> Actually no, because #95 was never a problem, since the ECR problem happens only when the qasm files are saved. The `get_one_benchmark` method is not saving the qasm files but just returning the Qiskit/TKET Circuit object. But nevertheless, this PR improves the postprocessing since it is now done automatically after the file is created.

Yeah, got that. I just meant that the issue can be closed (since it is not an issue after all).
 
> What do you mean by "linking"? Can you do it explicitly or is it done by just referencing somewhere to the issue?

On the right side of the PR page here under "Development" you can link any issues that are automatically closed once this PR gets merged. This can even be accomplished automatically when the PR description contains certain closing keywords. See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue for further explanations.

mqt/bench/utils/utils.py
nquetschlich: Done.

mqt/bench/utils/utils.py
nquetschlich: Done.

PR #165, opened 2023-02-04
Fix a typo in abstraction level selection block

jagandecapri:
I think the word is meant to be "rather". =)


burgholzer (approved):
Thanks for catching that!

PR #171, opened 2023-02-07
🐛 Quick-fix for uni-directional routing with TKET

### LINKED ISSUE #169: TKET does not respect uni-directional layout
At the moment, mapped circuits produced by TKET to architectures the are not bidirectional may contain invalid gates.
An example of this can be observed for 3-qubit DJ mapped to OQC Lucy with TKET using the Graph Placement, which results in 
```qasm
// Benchmark was created by MQT Bench on 2022-12-14
// For more information about MQT Bench, please visit https://www.cda.cit.tum.de/mqtbench/
// MQT Bench version: 0.2.2
// TKET version: 1.9.1
// Used Gate Set: ['rz', 'sx', 'x', 'ecr', 'measure']
// Coupling List: [[0, 1], [0, 7], [1, 2], [2, 3], [7, 6], [6, 5], [4, 3], [4, 5]]

OPENQASM 2.0;
include "qelib1.inc";
opaque ecr q0,q1;

qreg node[3];
creg c[2];
sx node[0];
rz(0.5*pi) node[1];
sx node[2];
rz(2.5*pi) node[0];
sx node[1];
rz(2.5*pi) node[2];
sx node[0];
rz(0.5*pi) node[1];
sx node[2];
rz(1.5*pi) node[0];
sx node[1];
rz(1.5*pi) node[2];
x node[0];
sx node[1];
x node[2];
rz(3.5*pi) node[0];
rz(3.5*pi) node[2];
ecr node[0],node[1];
sx node[0];
sx node[1];
rz(0.5*pi) node[0];
sx node[1];
ecr node[2],node[1];  # <--- invalid
rz(1.0*pi) node[1];
sx node[2];
rz(0.5*pi) node[2];
barrier node[0],node[2],node[1];
measure node[0] -> c[0];
measure node[2] -> c[1];
```
I marked the invalid gate. that direction should not be possible according to the coupling map.

~~I suspect this has something to do with the way the TKET Architecture is constructed.~~
The architecture is constructed just fine and only contains uni-directional links. Must be something else.

burgholzer:
Resolves #169.

⚡  also adds a new post-mapping, layout-preserving optimization pass for tket that should improve its results.

### DIFF
--- a/mqt/bench/tests/test_all.py
+++ b/mqt/bench/tests/test_all.py
@@ -240,6 +240,24 @@ def test_routing():
     assert qc.depth() > 0
 
 
+def test_unidirectional_coupling_map():
+    from pytket.architecture import Architecture
+
+    qc = get_benchmark(
+        benchmark_name="dj",
+        level="mapped",
+        circuit_size=3,
+        compiler="tket",
+        compiler_settings={"tket": {"placement": "graphplacement"}},
+        gate_set_name="oqc",
+        device_name="oqc_lucy",
+    )
+    # check that all gates in the circuit are in the coupling map
+    assert qc.valid_connectivity(
+        arch=Architecture(utils.get_cmap_oqc_lucy()), directed=True
+    )
+
+
 @pytest.mark.parametrize(
     "benchmark_name, level, circuit_size, benchmark_instance_name, compiler, compiler_settings, gate_set_name, device_name,",
     [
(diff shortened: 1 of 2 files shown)


burgholzer:
Hitting the GitHub API rate-limit again.. At some point we have to think about a solution where we use an OAuth token in the runners which would allow us to get up to 5000 request per hour.

PR #173, opened 2023-02-07
🔧 Configure GITHUB_TOKEN API Authorization

### LINKED ISSUE #172: GitHub API rate-limit
We are frequently hitting the GitHub API rate-limit again. GitHub limits unauthenticated API access to 60 requests per hour per IP address. With a personal access token this limit is increased to 5000. 

GitHub automatically provides such a token for each Action that is being run in the form of a `secrets.GITHUB_TOKEN`. This can be used to authenticate actions/runs.

_Originally posted by @burgholzer in https://github.com/cda-tum/MQTBench/issues/171#issuecomment-1421499254_

burgholzer:
Resolves #172.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -21,6 +21,8 @@ jobs:
         run: pip install -e .[coverage]
       - name: Generate Report
         run: pytest -v --cov=./ --cov-report=xml
+        env:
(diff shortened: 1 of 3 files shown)


burgholzer:
Reduced patch coverage here is fine as that's the part that only triggers when something doesn't quite work, which is precisely what this PR aims to mitigate.

PR #175, opened 2023-02-08
🚚 Move project to src layout

burgholzer:
This PR moves MQT Bench to a src layout similar to the MQT Predictor project.
Not only is this a common best practice for avoiding all kinds of packaging and testing problems (since `mqt.bench` is no longer also a top-level directory `mqt/bench`, but now `src/mqt/bench`), it also creates a much cleaner code structure, where source code is in `src` and tests are in `tests`.

Over the course of this transition, I also fixed some of the hard coded paths and replaced them with the corresponding `importlib.resource` code.

The actual changes are rather small and can be considered cleanup of existing code. Some modules have been moved around slighlty. Some modules got a more convenient include. A pytest and coverage configuration has been added.

Last but not least, this PR removes the deprecated `get_one_benchmark` function.


burgholzer:
The coverage changes here at first seemed weird. Then I noticed that the original coverage report is faulty itself. It shows many lines of important files as uncovered even though the respective lines are definitely covered by tests. 
I'll investigate further.

The actual coverage decrease here comes from the fact that many lines of code have been removed that were previously covered but were no longer needed.

nquetschlich (changes requested):
In general, all changes look very good. The only problem is, that starting the server via `mqt.bench` does not work locally for me. The benchmarks are downloaded, but the webpage is not reachable under the said url. 

Can you think of any reason for that?

Additionally, I get a `Segmentation fault` when the test `tests/test_bench.py::test_groundstate` is executed.

burgholzer:
Ah, might have missed changing the path. Thanks for catching that. I'll check later.

Tests run smoothly in the CI, so I'd guess that this is a macOS problem. At some point we should probably also run (some of) the tests on macOS and Windows in the CI.

burgholzer:
Ok, just checked again. Everything is running smoothly on my end. Have you tried it with a new virtual environment?

nquetschlich:
> Ok, just checked again. Everything is running smoothly on my end. Have you tried it with a new virtual environment?

I thought I had. Tested it again with new virtual environments for both python 3.9 and 3.10, it works now. I don't know why it has not worked before.

nquetschlich (approved):

PR #178, opened 2023-02-23
Evaluation Routine Implementation

nquetschlich:
This PR...
- adds the supermarq features (already used in the MQT Predictor) to this repository as well
- adds an `evaluation` module including visualization generation routines and functions to calculate different characteristics
- adds `nb-clean` to the pre-commit hooks


src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
+from __future__ import annotations
+
+import pickle
+import sys
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING or sys.version_info >= (3, 10, 0):  # pragma: no cover
+    pass
+else:
+    pass
```
burgholzer, line 10: seems like a leftover to me

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
...
+def create_statistics() -> None:
+    source_circuits_list = [file for file in Path(utils.get_qasm_output_path()).iterdir() if file.suffix == ".qasm"]
+
+    res = Parallel(n_jobs=-1, verbose=100)(delayed(evaluate_qasm_file)(filename) for filename in source_circuits_list)
+    target_dir = Path(resources.files("mqt.bench") / "evaluation/")
+    with Path(target_dir / "evaluation_data.pkl").open("wb") as f:
+        pickle.dump(res, f)
+
+
+def evaluate_qasm_file(filename: str) -> tuple[str, int, int, int, int, float, float, float, float, float]:
```
burgholzer, line 35: How about making that return type a `TypedDict` so that the individual evaluation metrics also get a meaning assigned to them?

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
...
+    import importlib_resources as resources
+
+
+def create_statistics() -> None:
+    source_circuits_list = [file for file in Path(utils.get_qasm_output_path()).iterdir() if file.suffix == ".qasm"]
+
+    res = Parallel(n_jobs=-1, verbose=100)(delayed(evaluate_qasm_file)(filename) for filename in source_circuits_list)
+    target_dir = Path(resources.files("mqt.bench") / "evaluation/")
+    with Path(target_dir / "evaluation_data.pkl").open("wb") as f:
+        pickle.dump(res, f)
```
burgholzer, line 32: I am not a big fan of dumping these results as part of the package since it needlessly increases the package size by 6.5MB. 
How long does the evaluation of these files take in general?
Is it feasible to just return `res` in this function?

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
...
+def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> Any:
+    offset = 0
+    for reg in qregs:
+        if qargs[index] not in reg:
+            offset += reg.size
+        else:
+            qubit_index = offset + reg.index(qargs[index])
+            return qubit_index
+    error_msg = "Qubit not found."
+    raise ValueError(error_msg)
```
burgholzer, line 63: Isn't that function already defined somewhere?
I vaguely remember seeing something like this. But maybe it was in the Predictor.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
...
+
+def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> Any:
+    offset = 0
+    for reg in qregs:
+        if qargs[index] not in reg:
+            offset += reg.size
+        else:
+            qubit_index = offset + reg.index(qargs[index])
+            return qubit_index
+    error_msg = "Qubit not found."
```
burgholzer, line 62: That error message could be a little more descriptive. As it is, it contains almost no information about the underlying problem.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,121 @@
...
+    assert 0 <= parallelism <= 1
+    assert 0 <= liveness <= 1
+
+    return (
+        program_communication,
+        critical_depth,
+        entanglement_ratio,
+        parallelism,
+        liveness,
+    )
```
burgholzer, line 121: Is that also "just" taken from the predictor?
If so, then consider moving it to an appropriate file here as well. IIRC, it's something like `reward` or so over there.
It certainly makes sense to move some common functions from the predictor over here.
Consequently, they should be removed from the Predictor in a subsequent update to avoid code duplication.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
```diff
@@ -0,0 +1,300 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "12d027e8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "with Path(\"./evaluation_data.pkl\").open(\"rb\") as f:\n",
+    "    eval_data = pickle.load(f)\n",
```
burgholzer, line 28: Already stated it above, but would it be too time consuming to perform the calculation of the evaluation data every time?
That would avoid having to store these pickle files.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
```diff
@@ -0,0 +1,300 @@
...
+    "    num_qubits,\n",
+    "    depth,\n",
+    "    num_all_gates,\n",
+    "    num_nonlocal_gates,\n",
+    "    program_communication,\n",
+    "    critical_depth,\n",
+    "    entanglement_ratio,\n",
+    "    parallelism,\n",
+    "    liveness,\n",
+    ") = [eval_data[:, i] for i in range(len(eval_data[0]))]\n",
```
burgholzer, line 42: I believe this code could be simplified by introducing the TypedDict as proposed above.
Because `eval_data` then is a simple list of these TypedDicts.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
```diff
@@ -0,0 +1,300 @@
...
+    "        if \"mapped_ibm_montreal\" in str_file:\n",
+    "            mapped_ibm_montreal_count += 1\n",
+    "        if \"mapped_ibm_washington\" in str_file:\n",
+    "            mapped_ibm_washington_count += 1\n",
+    "        if \"mapped_oqc_lucy\" in str_file:\n",
+    "            mapped_oqc_lucy_count += 1\n",
+    "        if \"mapped_rigetti\" in str_file:\n",
+    "            mapped_rigetti_count += 1\n",
+    "        if \"mapped_ionq\" in str_file:\n",
+    "            mapped_ionq_count += 1"
```
burgholzer, line 116: I'd say this "filename" to "counter dictionary" transformation could be packaged as a function in the evaluation module.
I'd put slightly more focus on making the method extensible with future architectures and gate-sets. Maybe TypedDicts or a slightly more sophisticated matching could help here as well.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
```diff
@@ -0,0 +1,300 @@
...
+   "source": [
+    "data = [perc_washington, perc_rigetti, perc_montreal, perc_ionq, perc_oqc]\n",
+    "colors = [\"#B7E6A5\", \"#7CCBA2\", \"#46AEA0\", \"#089099\", \"#00718B\"]\n",
+    "labels = [\"IBMQ Washington (127)\", \"Rigetti Aspen M2 (80)\", \"IBMQ Montreal (27)\", \"IonQ Harmony (11)\", \"OQC Lucy (8)\"]\n",
+    "pies = plt.pie(data, labels=labels, autopct=\"%1.1f%%\", colors=colors)\n",
+    "for pie in pies[0]:\n",
+    "    pie.set_alpha(0.6)\n",
+    "plt.savefig(\"results/pie.pdf\", format=\"pdf\", bbox_inches=\"tight\")\n",
+    "plt.show()"
+   ]
```
burgholzer, line 186: I just put this comment here because it is the first place where this suggestion fits, but what's coming next applies way more broadly to you repositories:

I am not a big fan of storing many pre-generated PDFs as part of our repositories as they quickly amount to a lot of files and just "lie around" in folders without much context.
Even more so, I am not a huge fan of Jupyter notebooks as they are really hard to read when not rendered properly by something like GitHub or your local IDE.

So the first thing I would advocate for, is moving as many utility functions as possible from the Jupyter notebook to a simple Python module. In this case here, that could be the `evaluation` module. Or it could be called `visualization` or something similar.
It would also be great if these visualisation methods could be somewhat covered with unit tests. That would probably require some kind of mocking/monkeypatching, but should be possible.
IMHO, the Jupyter notebooks should merely consist of some calls to library functions plus some added text for documentation purposes.

Now, I am a big fan of how we use Jupyter Notebooks in projects like QCEC and QMAP, where they are part of the documentation. See, e.g., https://github.com/cda-tum/qcec/blob/main/docs/source/Quickstart.ipynb, which gets rendered as https://qcec.readthedocs.io/en/latest/Quickstart.html in the documentation.
What happens here, is that the RtD documentation builder just executes the (clean) Jupyter notebooks and builds a documentation page out of that; including all the output.

All the evaluations we perform here (but also in the Predictor or the Problem Solver) would very nicely fit into their own documentation pages (assuming the generation of the respective images does not take hours).
This would eliminate the need for storing the individual PDFs in the repository as they would just be generated on-the-fly. In addition, documentation pages provide quite a lot more context than the individual PDF files lying around.

I guess what I am trying to say is: We should really get some RtD up and running cause it could significantly improve the way we can visualize data and results.

tests/test_bench.py
```diff
@@ -763,3 +763,22 @@ def test_oqc_postprocessing():
...
+    filename = "test_5.qasm"
+    qc.qasm(filename=filename)
+    path = Path(filename)
+    res = evaluation.evaluate_qasm_file(filename)
+    assert len(res) == NUM_EVAL_ATTRIBUTES
+    assert res[0] == filename
+    for elem in res[1:5]:
+        assert isinstance(elem, int)
+    for elem in res[5:]:
+        assert isinstance(elem, float)
```
burgholzer, line 782: I'd say most of these assertions become redundant when the evaluation returns a TypedDict.
These tests could be a little more comprehensive I guess.

burgholzer (changes requested):
Thanks for this PR!
As (almost) always, I have a couple of thoughts on this or it triggered some thoughts.
These thoughts are best summarised in the comments below, but I think the gist of it is:

I believe there is merit in having these kind of evaluations and visualisations as part of the repository. However, I feel that the current way is not really sustainable when thinking about the future and needs to be automated to be practical. Such an automation could happen in the form of proper RtD documentation and some generalisations of the functions used to generate the visualisations.

I do think that these documentation changes deserve their own PR though and the PR here can be merged as an intermediate step, after all the other comments have been addressed.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Unfortunately, the generation takes around 11h. Therefore returning `res` might not be a viable approach.

burgholzer: Ok. Alternative suggestion: Can we dynamically download the `evaluation_data.pkl` from release assets just like we do with the `.zip`?

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Removed that fragment.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Good point, implemented such a `TypedDict`.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Yes, removed it according to the MQT Predictor and will delete it there in the next days.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Yes, removed it according to the MQT Predictor and will delete it there in the next days.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Improved on that.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
nquetschlich: Improved that.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
nquetschlich: I agree with you. The documentation improvement and shift to Read-the-Docs is on my todo-list for the next weeks.

src/mqt/bench/evaluation/evaluation.py
burgholzer: Sounds good. Keep in mind that a new MQT Bench release is necessary for the functions to be available in the Predictor and that the lower version cap of MQT in the Predictor needs to be increased accordingly.

src/mqt/bench/utils.py
```diff
@@ -342,3 +343,73 @@ def postprocess_single_oqc_file(filename: str):
 
 def create_zip_file():
     return subprocess.call(f"zip -rj {get_zip_file_path()} {get_qasm_output_path()}", shell=True)
+
+
+def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> Any:
```
burgholzer, line 348: Why are you annotating the return type with `Any` here? Shouldn't that just be `int`?

src/mqt/bench/utils.py
```diff
@@ -342,3 +343,73 @@ def postprocess_single_oqc_file(filename: str):
...
+        else:
+            qubit_index = offset + reg.index(qargs[index])
+            return qubit_index
+    error_msg = "Global qubit index for given local qubit index not found."
+    raise ValueError(error_msg)
+
+
+def calc_supermarq_features(
+    qc: QuantumCircuit,
+) -> tuple[float, float, float, float, float]:
```
burgholzer, line 362: Also a nice opportunity to introduce a TypedDict, or probably even better: a dataclass.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -25,97 +17,43 @@
...
+    filename: str
+    num_qubits: int
+    depth: int
+    num_gates: int
+    num_multiple_qubit_gates: int
+    program_communication: float
+    critical_depth: float
+    entanglement_ratio: float
+    parallelism: float
+    liveness: float
```
burgholzer, line 38: Now that I think of it, a `dataclass` is probably more suitable here since the dictionary just creates an unnecessary overhead compared to the `tuple`.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Good points. I created an issue (https://github.com/cda-tum/MQTPredictor/issues/74) in the Predictor to be reminded of that.

src/mqt/bench/evaluation/evaluation_visualization.ipynb
nquetschlich: I added two counting methods in the `evaluation.py` file to remove most of the calculations (but simple divisions ) from the jupyter notebook

src/mqt/bench/utils.py
nquetschlich: Fixed that.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -57,3 +56,12 @@ def evaluate_qasm_file(filename: str) -> EvaluationResult:
         "parallelism": parallelism,
         "liveness": liveness,
     }
+    return cast(EvaluationResult, res_dict)
```
burgholzer, line 59: No need to cast here. Just do
```suggestion
    return EvaluationResult(
        filename=filename,
        num_qubits=int(str(filename).split("_")[-1].split(".")[0]),
        depth=qc.depth(),
        etc.
    )
```
And at that point, you can use a data class as well, which would look exactly the same here.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -57,3 +56,12 @@ def evaluate_qasm_file(filename: str) -> EvaluationResult:
...
+
+
+def count_occurrences(filenames: list[str], search_str: str) -> int:
+    return sum([search_str in filename for filename in filenames])
+
+
+def count_qubit_numbers_per_compiler(filenames: list[str], compiler: str) -> list[int]:
+    return [int(str(filename).split("_")[-1].split(".")[0]) for filename in filenames if compiler in filename]
```
burgholzer, line 67: These functions aren't covered by tests. That should be fairly easy to change.

src/mqt/bench/utils.py
```diff
@@ -345,7 +345,7 @@ def create_zip_file():
     return subprocess.call(f"zip -rj {get_zip_file_path()} {get_qasm_output_path()}", shell=True)
 
 
-def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> Any:
+def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> int:
```
burgholzer, line 348: Coming back to this again: Why is the `qargs` argument typed as `list[Any]` instead of a proper type?

tests/test_bench.py
```diff
@@ -774,11 +774,7 @@ def test_evaluate_qasm_file():
     qc.qasm(filename=filename)
     path = Path(filename)
     res = evaluation.evaluate_qasm_file(filename)
-    assert len(res) == NUM_EVAL_ATTRIBUTES
-    assert res[0] == filename
-    for elem in res[1:5]:
-        assert isinstance(elem, int)
-    for elem in res[5:]:
-        assert isinstance(elem, float)
+    assert type(res) == dict
```
burgholzer, line 777: the type isn't only a dict. it has to be the typed dict you created.

burgholzer (changes requested):
Just some further comments here because I had some time to go through this.

src/mqt/bench/utils.py
```diff
@@ -342,3 +343,74 @@ def postprocess_single_oqc_file(filename: str):
...
+
+def calc_qubit_index(qargs: list[Any], qregs: list[QuantumRegister], index: int) -> int:
+    offset = 0
+    for reg in qregs:
+        if qargs[index] not in reg:
+            offset += reg.size
+        else:
+            qubit_index = offset + reg.index(qargs[index])
+            return qubit_index
+    error_msg = "Global qubit index for given local qubit index not found."
```
burgholzer, line 356: While this error message is better than the message before, it still contains absolutely no information about the local qubit index that was the cause of the error.
```suggestion
    error_msg = f"Global qubit index for local qubit {index} index not found."
```

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Implemented it now as a dataclass

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Adjusted the code accordingly.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Added respective tests.

tests/test_bench.py
nquetschlich: Changed that.

src/mqt/bench/utils.py
nquetschlich: tbh, because I have overlooked that. Added it by now.

src/mqt/bench/utils.py
nquetschlich: Implemented it.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@dataclass
+class EvaluationResult:
+    filename: str
+    num_qubits: int
+    depth: int
+    num_gates: int
+    num_multiple_qubit_gates: int
+    supermarq_features: dict[str, float]
```
burgholzer, line 36: Wrong type ;o)
We really need mypy here as well.

nquetschlich: Fixed that...

src/mqt/bench/evaluation/evaluation.py
nquetschlich: I removed the file from the repo as discussed.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+    filename: str
+    num_qubits: int
+    depth: int
+    num_gates: int
+    num_multiple_qubit_gates: int
+    supermarq_features: utils.SupermarqFeatures
+
+
+def evaluate_qasm_file(filename: str) -> EvaluationResult:
+    print(filename)
```
burgholzer: I guess this is still some leftover from debugging?
This should probably use logging instead of print. Fine for this PR, but maybe to keep in mind.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+    qc = QuantumCircuit.from_qasm_file(filename)
+    supermarq_features = utils.calc_supermarq_features(qc)
```
burgholzer, line 42: Can just be inlined below.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,61 @@
...
+    import importlib_resources as resources
+
+
+def create_statistics() -> None:
+    source_circuits_list = [file for file in Path(utils.get_qasm_output_path()).iterdir() if file.suffix == ".qasm"]
+    res_dicts = Parallel(n_jobs=-1, verbose=100)(
+        delayed(evaluate_qasm_file)(str(filename)) for filename in source_circuits_list
+    )
+    target_dir = Path(resources.files("mqt.bench") / "evaluation/")
+    with Path(target_dir / "evaluation_data.pkl").open("wb") as f:
```
burgholzer, line 25:
```suggestion
    with (resources.files("mqt.bench") / "evaluation" / "evaluation_data.pkl").open("wb") as f:
```
Should also work and be more compact.

src/mqt/bench/utils.py
```diff
@@ -342,3 +358,74 @@ def postprocess_single_oqc_file(filename: str):
...
+        all_indices = [first_qubit]
+        if len(qargs) == 2:
+            second_qubit = calc_qubit_index(qargs, qc.qregs, 1)
+            all_indices.append(second_qubit)
+        for qubit_index in all_indices:
+            to_be_added_entries = all_indices.copy()
+            to_be_added_entries.remove(int(qubit_index))
+            connectivity_collection[int(qubit_index)].extend(to_be_added_entries)
+
+    connectivity: list[Any] = []
```
burgholzer, line 397: Needs a proper type ;o)
Shouldn't be too hard to infer.

tests/test_bench.py
```diff
@@ -763,3 +763,68 @@ def test_oqc_postprocessing():
     )
     assert QuantumCircuit.from_qasm_file(str(path))
     path.unlink()
+
+
+NUM_EVAL_ATTRIBUTES = 10
```
burgholzer, line 768: Seems like this is not needed anymore

burgholzer (changes requested):
A last round of review before this can be merged. Just minor details.

src/mqt/bench/evaluation/evaluation.py
```diff
@@ -0,0 +1,61 @@
...
+
+def count_occurrences(filenames: list[str], search_str: str) -> int:
+    return sum([search_str in filename for filename in filenames])
+
+
+def count_qubit_numbers_per_compiler(filenames: list[str], compiler: str) -> list[int]:
+    return [int(str(filename).split("_")[-1].split(".")[0]) for filename in filenames if compiler in filename]
+
+
+create_statistics()
```
burgholzer, line 61: What is this lonely call doing here? 🙃

src/mqt/bench/evaluation/evaluation.py
nquetschlich: True, removed that.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: Done.

tests/test_bench.py
nquetschlich: Removed it.

src/mqt/bench/evaluation/evaluation.py
nquetschlich: That happened accidentally.

src/mqt/bench/utils.py
nquetschlich: Yes, was quite simple. Additionally, I improved the `connectivity` initialization routine below.Yes, was quite simple. Additionally, I improved the `connectivity` initialization routine below.

burgholzer (approved):
LGTM 👍🏼

PR #179, opened 2023-02-23
Fix Typos

nquetschlich:
Fixed typos:
- Renamed "IonQ" device to "IonQ Harmony" to be consistent with the paper 
- Removed ":" in Level Names to be consistent with the paper 
- Changed "Native-Gates" to "Native Gates" to be consistent with the paper

### DIFF
--- a/src/mqt/benchviewer/templates/description.html
+++ b/src/mqt/benchviewer/templates/description.html
@@ -82,7 +82,7 @@
           representations of the circuits from the algorithmic level.
         </li>
         <li>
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
LGTM 👍🏼 

When we really start working on the more generic Device and Provider Integration we should also try to keep these names consistent and establish a naming scheme for all the different providers and devices. Ideally, when the provider has a certain `provider_name`, and the devices have a certain `device_name`, calibration and hardware information can be found in a file `provider_name_device_name`. (That's at least close to what I tried in that Predictor PR)

PR #183, opened 2023-02-24
Fix Server Deploy Workflow

nquetschlich:


.github/workflows/server_deploy.yml
```diff
@@ -1,8 +1,11 @@
 # Based on: https://zartman.xyz/blog/gh-static-deploy/
 name: Server Deployment
 on:
-  release:
-    types: [published]
+  workflow_run:
+    workflows: ["Deploy to PyPI"]
+    branches: [main]
+    types:
+      - completed
```
burgholzer, line 8: Doesn't the workflow run also complete when the package is not pushed to PyPI?
I am not quite sure that this works as intended.

nquetschlich: True, good catch. I added the same if statement from the `Deploy to PyPI` workflow.

burgholzer (approved):
I believe this should work now.

Issue #167, opened 2023-02-07
Qubit Allocation for Mapped Layer using tket


burgholzer:
I just stumbled upon this: `tket` doesn't perform any kind of full ancilla allocation as Qiskit does when mapping quantum circuits to architectures. As a result, the mapped circuits resulting from tket contain quantum registers with varying size; depending on how the circuit was routed. For example, if the largest qubit index being used in the circuit is `32`, then the circuit mapped to the 127-qubit `ibm_washington` architecture will still only have a `qreg node[32];` definition.

While this is not urgent, I found this surprising and believe it would be good to resolve this disparity between the files generated by qiskit and the files generated by tket.

nquetschlich:
see https://github.com/CQCL/tket/issues/770, https://github.com/CQCL/tket/issues/774, and https://cqcl.github.io/tket/pytket/api/circuit_class.html#pytket.circuit.Circuit.add_blank_wires.

PR #191, opened 2023-03-08
🩹 Fix of Circuit Size for TKET on Mapped Level

nquetschlich:
This PR solves #167.

As soon as there is a new `pytket` version, some of the induced compilation steps can be subsumed by one `PeepHoleOptimise2Q` compilation pass call. Currently, this cannot be done since this pass can destroy the mapping (see https://github.com/CQCL/tket/issues/776 for more details).

### DIFF
--- a/src/mqt/bench/tket_helper.py
+++ b/src/mqt/bench/tket_helper.py
@@ -220,9 +220,6 @@ def get_mapped_level(
         print("TKET Exception Mapped: ", e)
         return False
 
-    diff = max(list(map(max, cmap))) + 1 - qc_tket.n_qubits  # offset of one is added because the indices start at 0
-    qc_tket.add_blank_wires(diff)
-
     native_gatenames = get_rebase(gate_set_name, True)
     native_gate_set_rebase = get_rebase(gate_set_name)
     arch = architecture.Architecture(cmap)
@@ -233,10 +230,13 @@ def get_mapped_level(
     PlacementPass(placer).apply(qc_tket)
     RoutingPass(arch).apply(qc_tket)
     FullPeepholeOptimise(target_2qb_gate=OpType.TK2, allow_swaps=False).apply(qc_tket)
-    native_gate_set_rebase.apply(qc_tket)
     if not qc_tket.valid_connectivity(arch, directed=True):
         CXMappingPass(arc=arch, placer=placer, directed_cx=True, delay_measures=False).apply(qc_tket)
     native_gate_set_rebase.apply(qc_tket)
+
+    highest_used_qubit_index = max(list(map(max, cmap)))
+    diff = highest_used_qubit_index + 1 - qc_tket.n_qubits  # offset of one is added because the indices start at 0
+    qc_tket.add_blank_wires(diff)
 
     if return_qc:
         return qc_tket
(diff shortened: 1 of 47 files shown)


burgholzer:
Just to double check: we are still waiting on some kind of tket fix here, right?

nquetschlich:
> Just to double check: we are still waiting on some kind of tket fix here, right?

Yes, basically we are waiting for a fix to https://github.com/CQCL/tket/issues/777.

nquetschlich:
> > Just to double check: we are still waiting on some kind of tket fix here, right?
> 
> Yes, basically we are waiting for a fix for [CQCL/tket#777](https://github.com/CQCL/tket/issues/777).

Unfortunately, the simple solution mentioned in the PR does not work in our case, since apparently the `CXMappingPass` compilation pass does not properly work in combination with the `add_blank_wires`.

nquetschlich:
> > > Just to double check: we are still waiting on some kind of tket fix here, right?
> > 
> > 
> > Yes, basically we are waiting for a fix for [CQCL/tket#777](https://github.com/CQCL/tket/issues/777).
> 
> Unfortunately, the simple solution mentioned in the PR does not work in our case, since apparently the `CXMappingPass` compilation pass does not properly work in combination with the `add_blank_wires`.

We are currently waiting for a solution for https://github.com/CQCL/tket/issues/776, but already got clarifications for https://github.com/CQCL/tket/issues/777.

src/mqt/bench/tket_helper.py
```diff
@@ -310,16 +311,21 @@ def get_mapped_level(
...
     native_gate_set_rebase.apply(qc_tket)
     FullPeepholeOptimise(target_2qb_gate=OpType.TK2).apply(qc_tket)
     placer = LinePlacement(arch) if lineplacement else GraphPlacement(arch)
-    PlacementPass(placer).apply(qc_tket)
-    RoutingPass(arch).apply(qc_tket)
-    FullPeepholeOptimise(target_2qb_gate=OpType.TK2, allow_swaps=False).apply(qc_tket)
-    native_gate_set_rebase.apply(qc_tket)
-    if not qc_tket.valid_connectivity(arch, directed=True):
-        CXMappingPass(arc=arch, placer=placer, directed_cx=True, delay_measures=False).apply(qc_tket)
+    CXMappingPass(arc=arch, placer=placer, directed_cx=True, delay_measures=False).apply(qc_tket)
```
burgholzer, line 322: What's the reason for this pass to now be applied unconditionally? Shouldn't this just be applied when the connectivity is not satisfied after the standard pass?

Asked differently: shouldn't this special treatment only apply for directed architectures?

burgholzer (approved):
Just one comment where I am not 100% about. Otherwise LGTM.

src/mqt/bench/tket_helper.py
nquetschlich: I thought always using the `CXMappingPass` would remove the `if` condition and, by that, simplify the code. Do you see a disadvantage in using that MappingPass as the default one?

burgholzer: Yeah. I would guess that the default mapping pass is superior to the `CXMappingPass` otherwise the `CXMappingPass` would be used per default in TKET.

nquetschlich: Hm, I see. To be honest I thought the `CXMappingPass` would be more powerful (maybe I was misled because `CXMappingPass` sounded more sophisticated to me than `RoutingPass`). I adapted the code accordingly.

burgholzer (approved):
LGTM. Maybe you could expand the PR description and title a little so it also includes information on the fix for the circuit directionality.

BTW: we should probably create a new benchmark suite version at some point.

PR #197, opened 2023-03-28
🐛 SupermarQ Feature Calculation

nquetschlich:
So far, the Qiskit.QuantumCircuit object is not copied and was modified by the method. To avoid undesired behavior (such as the compiled circuit not having measurements anymore), the QuantumCircuit is now copied before.

### DIFF
--- a/src/mqt/bench/utils.py
+++ b/src/mqt/bench/utils.py
@@ -16,7 +16,6 @@ import numpy as np
 from pytket import __version__ as __tket_version__
 from qiskit import QuantumCircuit, __qiskit_version__
 from qiskit.providers.fake_provider import FakeMontreal, FakeWashington
-from qiskit.transpiler.passes import RemoveBarriers
 from qiskit_optimization.applications import Maxcut
 
 if TYPE_CHECKING or sys.version_info >= (3, 10, 0):  # pragma: no cover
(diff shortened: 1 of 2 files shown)


src/mqt/bench/utils.py
```diff
@@ -313,7 +313,7 @@ def calc_qubit_index(qargs: list[Qubit], qregs: list[QuantumRegister], index: in
 def calc_supermarq_features(
     qc: QuantumCircuit,
 ) -> SupermarqFeatures:
-    qc.remove_final_measurements(inplace=True)
+    qc = qc.remove_final_measurements(inplace=False)
```
burgholzer, line 316: Instead of removing barriers and measurements here (and now creating a copy of the circuit), you could just add an `if` statement in the loop over `qc.data` (L323) that skips over measurements and barriers. That would not alter the circuit and would not require a copy.
(You have to change one of the `_` to an actual loop index to get the name of the instruction)

burgholzer (changes requested):
I think this can be solved in a better way without requiring a circuit copy, see below.

src/mqt/bench/utils.py
nquetschlich: True, thanks for pointing that out. I modified the code accordingly. For that, I had to make minor modifications for the number of gates and depths calculations.

src/mqt/bench/utils.py
```diff
@@ -313,38 +312,40 @@ def calc_qubit_index(qargs: list[Qubit], qregs: list[QuantumRegister], index: in
...
-        all_indices = [first_qubit]
-        if len(qargs) == 2:
-            second_qubit = calc_qubit_index(qargs, qc.qregs, 1)
-            all_indices.append(second_qubit)
-        for qubit_index in all_indices:
-            to_be_added_entries = all_indices.copy()
-            to_be_added_entries.remove(int(qubit_index))
-            connectivity_collection[int(qubit_index)].extend(to_be_added_entries)
+    for instruction, qargs, _ in qc.data:
+        if instruction.name != "barrier" and instruction.name != "measure":
```
burgholzer, line 321: Small nit: I would prefer a 
```python
if instruction.name == "barrier" or instruction.name == "measure":
  continue
```
here to reduce the indentation of the rest of the code block.

src/mqt/bench/utils.py
```diff
@@ -313,38 +312,40 @@ def calc_qubit_index(qargs: list[Qubit], qregs: list[QuantumRegister], index: in
...
-    depth = qc.depth()
+    depth = qc.depth(lambda x: x[0].name != "barrier" and x[0].name != "measure")
     program_communication = np.sum(connectivity) / (qc.num_qubits * (qc.num_qubits - 1))
 
     if num_multiple_qubit_gates == 0:
         critical_depth = 0.0
     else:
-        critical_depth = qc.depth(filter_function=lambda x: len(x[1]) > 1) / num_multiple_qubit_gates
+        critical_depth = (
+            qc.depth(filter_function=lambda x: len(x[1]) > 1 and x[0].name != "barrier") / num_multiple_qubit_gates
```
burgholzer, line 347: Why does this not exclude measurements?

burgholzer (changes requested):
Thanks for the changes. Looks better. Two small comments though.

src/mqt/bench/utils.py
nquetschlich: Because all local gates operating only on a single qubit are already filtered by the first condition `lambda x: len(x[1]) > 1`. Do you see a benefit in explicitly list the measurements as well for clarity?

src/mqt/bench/utils.py
nquetschlich: Implemented that.

src/mqt/bench/utils.py
burgholzer: Ah. Missed that. Then consider this resolved 😎

PR #214, opened 2023-06-26
❇️  v1.0.0 update

nquetschlich:
This PR...

- adds an additional IonQ Device: Aria 1 with 25 Qubits
- adds an additional native gate set from Quantumnuum
- adds an additional Quantinuum Device: H2 with 32 Qubits
- updates both Qiskit's and TKET's version to the latest one
- fixed the value domain for random parameters for certain benchmarks
- adds two new benchmark algorithms: Random Circuits and Quantum Neural Networks
- removes two benchmarks (temporarily): QGAN and HHL


nquetschlich:
Currently, a few tests are failing because the 11-qubit IonQ device has been renamed to IonQ Harmony and the old database with the old name is downloaded since the new database will be part of the v1.0.0 release.

src/mqt/bench/benchmarks/random.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    qc = random_circuit(num_qubits, num_qubits * 2, measure=False, seed=10)
+    gates = list(set(utils.get_openqasm_gates()) - {"rccx", "csx", "cu"})
+    qc = transpile(
+        qc,
+        basis_gates=gates,
+        seed_transpiler=10,
+        optimization_level=1,
+    )
+    qc.measure_all()
```
burgholzer, line 20:
```suggestion
    qc = random_circuit(num_qubits, num_qubits * 2, measure=True, seed=10)
```
Why not just?

src/mqt/bench/benchmarks/shor.py
```diff
@@ -32,3 +52,222 @@ def get_instance(choice: str) -> list[int]:
...
+    def quantum_instance(self) -> QuantumInstance | None:
+        """Returns quantum instance."""
+        return self._quantum_instance
+
+    @quantum_instance.setter
+    def quantum_instance(self, quantum_instance: QuantumInstance | Backend) -> None:
+        """Sets quantum instance."""
+        if isinstance(quantum_instance, Backend):
+            quantum_instance = QuantumInstance(quantum_instance)
+        self._quantum_instance = quantum_instance
```
burgholzer, line 78: I think this can be safely removed. In general it would be good to clean this file up a little and minimize it. Although, I totaly get, that you probably don't want to do this as part of this PR.
```suggestion
```

src/mqt/bench/benchmark_generator.py
```diff
@@ -283,8 +267,8 @@ def get_benchmark(  # noqa: PLR0911, PLR0912, PLR0915
     benchmark_instance_name -- Input selection for some benchmarks, namely "groundstate" and "shor"
     compiler -- "qiskit" or "tket"
     CompilerSettings -- Data class containing the respective compiler settings for the specified compiler (e.g., optimization level for Qiskit or placement for TKET)
-    gate_set_name -- "ibm", "rigetti", "ionq", or "oqc"
-    device_name -- "ibm_washington", "ibm_montreal", "rigetti_aspen_m2", "ionq11", ""oqc_lucy""
+    gate_set_name -- "ibm", "rigetti", "ionq", "oqc", or "quantinuum"
+    device_name -- "ibm_washington", "ibm_montreal", "rigetti_aspen_m2", "ionq_harmony", "ionq_aria1", ""oqc_lucy"", ""quantinuum_lucy"",
```
burgholzer, line 271:
```suggestion
    device_name -- "ibm_washington", "ibm_montreal", "rigetti_aspen_m2", "ionq_harmony", "ionq_aria1", ""oqc_lucy"", ""quantinuum_h2"",
```

C&P mistake ;o)

burgholzer (changes requested):
Just had a quick look through this PR. Overall it looks good and I like that some of the methods are finally becoming a little less repetitive and generic.
Just some small comments that I believe should be addressed.
After that, "Go" from my end.

src/mqt/bench/benchmarks/random.py
nquetschlich: There are two reasons for that. 
1) The `measure=True` does not add a barrier in front of the measurements. Could be neglected, but all other benchmark circuits have such a barrier and I wanted it to be consistent.
2) The transformation from a `Qiskit.QuantumCircuit` to a `pytket.Circuit` via the `pytket-qiskit` transformation functions crashes if those mentioned gates are used why I removed them using that transpilation pass. I am aware that this is not ideal.

burgholzer: I get the consistency argument. +1 on that!

The issue with the gates is non-ideal. Would you mind creating an issue in the tket repository for that?

nquetschlich: Got point, I created the issue [here](https://github.com/CQCL/pytket-qiskit/issues/138).

burgholzer (approved):

PR #283, opened 2024-01-19
🔧  add all-to-all connectivity for ionq aria device

flowerthrower:
Configuration file only had one direction coupling map specified.
I added the other direction. Noticed this through a failing test in MQTPredictor.


nquetschlich:
@flowerthrower Thank you very much. Just to double-check: we do need that fix for IonQ Harmony?

PR #286, opened 2024-01-22
Integrating BQSKit into MQTBench benchmark generation module

jagandecapri:
This PR addresses https://github.com/cda-tum/mqt-bench/issues/282. In summary, a new module called `bsqkit_helper` is created to integrate [BSQKit](https://bqskit.readthedocs.io/en/latest/index.html) into MQTBench. There are some open issues. Namely, QASM generation of the benchmark is not working at the moment and the test cases currently doesn’t test for QASM generation. There are also other bugs as highlighted below. Numpy was also used in the module which is open for discussion as I saw Numpy is not a dependency in MQTBench currently. Test coverage might be lower as I didn’t add new test cases for the compiler cache and custom gates (ECR and XXPlusYY). Transpilation to only QASM 2.0 gates using Qiskit is needed before converting to bqskit circuit as not all Qiskit circuit gates that is used in the benchmark has an equivalent in bqskit, e.g: `mcx` gate. Without the transpilation, bsqkit converts the unsupported gate to arbitrarily large block of custom gate which then fails to compile. Runtime for test cases seems to have increased quite considerably with the introduction of bqskit. More details as below:

### QASM string representation

- Issue https://github.com/BQSKit/bqskit/issues/214
- QASM string conversion from a bqskit circuit is not working well. It breaks when converting certain control gates
- Currently,  test cases for bqskit is not testing the QASM string generation

### `cry` gate bug

- Issue https://github.com/BQSKit/bqskit/issues/217. Due to this, native and mapped level test cases which has `cry` gate(s) execute as following:
    - `pricingcall` test case executes indefinitely
    - `pricingput` and `random` test case sometimes works and sometimes face the same problem of indefinite execution as `pricingcall`. For `random`, since it is a random circuit from Qiskit, I suspect when there are multiple `cry` gates, the test cases starts to go into indefinite execution

### Custom gates

- Implemented two custom gates, ECR and XXPlusYY, following Qiskit little-endian convention. The former is supported natively by OQC Lucy and the latter by Rigetti.
    - ECR Gate
        - Type check is disabled for inherited bqskit classes as mypy throws error otherwise. It checks for type information from the inherited classes which doesn’t support type checks I think. Not sure whether this is the right fix.
        - Need verification whether unitary implemented is correct. Obtained it from Qiskit documentation for [ECR Gate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.ECRGate)
        - Created issue in bqskit to add this gate in their library https://github.com/BQSKit/bqskit/issues/216
    - XXPlusYY Gate
        - Largely inspired by bqskit [FSIM Gate](https://bqskit.readthedocs.io/en/latest/_modules/bqskit/ir/gates/parameterized/fsim.html#FSIMGate)
        - Type check is disabled for inherited bqskit classes as mypy throws error otherwise. It checks for type information from the inherited classes which doesn’t support type checks I think. Not sure whether this is the right fix.
        - There are two types of optimization algorithm that is used by bqskit for parameterized gates. They are `DifferentiableUnitary` and `LocallyOptimizableUnitary`. A `DifferentiableUnitary` can be used in instantiation calls that use a gradient based optimization subroutine. On the other hand,`LocallyOptimizableUnitary`can be used in instantiation calls that use a tensor-network based approach like QFactor. More info at [here](https://github.com/BQSKit/bqskit/blob/1ee59f11da206c3b18667c7691aded816016c8ed/examples/tutorials/Introduction%20to%20BQSKit%20IR.ipynb#L18)
        - The gate implements only the function for differentiable unitary optimization. From my understanding reading the FSIM and [U3](https://bqskit.readthedocs.io/en/latest/_modules/bqskit/ir/gates/parameterized/u3.html#U3Gate ) gate implementation, differentiable unitary returns a numpy ndarray containing the partial differentation of the gate’s unitary matrix with respect to each parameter which is what I have implemented. On the other hand, for locally optimizable unitary, I’m not really sure what is the right implementation.
        - Need verification whether unitary in `get_unitary` and `get_grad` is correct. Obtained unitary from Qiskit documentation for [XXPlusYY](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.XXPlusYYGate) gate. As for the gradient, I derived manually and verified with WolframAlpha.
        - This gate uses numpy library in the `get_grad` function. However, numpy is not listed as dependency in `pyproject.toml`. I think it gets installed with other library. Should we make it a dependency in MQTBench? if we agree on using numpy, then, some functionality that is using `cmath` and `math` module in `bsqkit_helper` can be replaced with numpy
        - Issue could be created in bsqkit if they are interested to add this gate to their library

### CPhase gate
- Added CPhase gate for Rigetti native gates

### Cache for compiler

- Each call to the `compile` function creates resources to parallelize the compilation across the processors in the system. Hence, there is a waiting time for the setup to occur. Bqskit documentation suggested caching the compiler instance for repeated call of the `compile` function which is implemented in the `CacheCompiler` class using the singleton design pattern
- The compiler instance is registed with the `atexit` module to close it when the script exits
- Type check is suppressed for `__new__` method as `ruff` was throwing error for the return type which seems to be correct from my point of view

## Runtime of test case

- I timed the pytest command `time pytest -k 'test_quantumcircuit_native_and_mapped_levels'`  for only bqskit, minus the circuits mentioned in the `cry` bug section above. It takes **~54 minutes** to run this one test function in my laptop. The other test functions execution time seems reasonable

## Code structure

The `bsqkit_helper` class seems to be doing a lot of things, managing cache for the compiler, creating custom gates and benchmark generation. Maybe it could be broken to different files. I didn’t want to change the structure of the repository, so I’ve put all of it in the helper class. Open for discussion. 

With that, looking forward to your kind review and discussions. 🙂


jagandecapri:
Fixed failing test cases due to bugs in code. Some test cases were left to failed intentionally to indicate the part that needs to be updated when the bugs in bqskit library are fixed. Alternatively, in some cases, I have left TODOs comments in `test_bench.py` to indicate the parts that needs updating once the bugs in bqskit library are fixed.

src/mqt/bench/bqskit_helper.py
```diff
@@ -0,0 +1,453 @@
...
+        "rz": RZGate(),
+        "rxx": RXXGate(),
+        "rzz": RZZGate(),
+        "sx": SXGate(),
+        "x": XGate(),
+        "cp": CPGate(),
+        "cx": CNOTGate(),
+        "cz": CZGate(),
+        "ecr": ECRGate(),
+        "xx_plus_yy": XXPlusYYGate(),
```
nquetschlich, line 182: Could you also make sure that this gate is implemented for TKET as well in the `tket_helper.py` file? Same for the `cp` gate.

src/mqt/bench/bqskit_helper.py
```diff
@@ -0,0 +1,453 @@
...
+            qc,
+            basis_gates=gates,
+            seed_transpiler=10,
+            optimization_level=0,
+        )
+        qc_bqskit = qiskit_to_bqskit(qc)
+    except Exception as e:
+        print("BQSKit Exception Indep: ", e)
+        return False
+
```
nquetschlich, line 250: Where is the BQSKit compilation happening? I don't see it here.

src/mqt/bench/bqskit_helper.py
```diff
@@ -0,0 +1,453 @@
...
+            basis_gates=gates,
+            seed_transpiler=10,
+            optimization_level=0,
+        )
+        qc_bqskit = qiskit_to_bqskit(qc)
+    except Exception as e:
+        print("BQSKit Exception NativeGates: ", e)
+        return False
+
+    gate_set = provider.get_native_gates()
```
nquetschlich, line 338: all those compilation commands should be within the `try` block

tests/test_bench.py
```diff
@@ -70,15 +73,18 @@ def sample_filenames() -> list[str]:
         "ae_indep_qiskit_10.qasm",
         "ghz_nativegates_rigetti_qiskit_opt3_54.qasm",
         "ae_indep_tket_93.qasm",
+        "ae_indep_bqskit_93.qasm",
```
nquetschlich, line 76: Maybe a smaller circuit would be better suited for BQSKit because of their runtime (especially for optimization levels >2)

tests/test_bench.py
```diff
@@ -1031,14 +1224,15 @@ def test_evaluate_qasm_file() -> None:
...
     ("search_str", "expected_val"),
     [
+        ("bqskit", 3),
         ("qiskit", 10),
         ("tket", 3),
         ("nativegates", 2),
-        ("indep", 2),
-        ("mapped", 9),
+        ("indep", 3),
+        ("mapped", 11),
```
nquetschlich, line 1232: Why have the test case sizes changed?

nquetschlich (changes requested):
Thanks a lot for your effort and this PR, @jagandecapri. I am very happy that we will (soon) also support BQSKit as another compiler. I have just mentioned a few things that caught my attention. Additionally I was wondering, whether we really need all the newly introduced type ignores. Please carefully double check, whether those are actually necessary.

jagandecapri:
- [ ] To add bqskit version to benchmark files https://github.com/jagandecapri/MQTBench/blob/918a3c0f2eea84ab935d896513b322849a81c3e7/src/mqt/bench/utils.py#L197-L200

nquetschlich:
No integration is planned in the near future.

PR #295, opened 2024-02-06
Minor Webpage Improvements

nquetschlich:
This PR
- updates the favicon (previously drawn bell state quantum circuit, now MQT logo)
- updates the native gates section of the webpage to also show the considered gates per gate set (excluding `id`, `barrier`, `measurement` gates)


src/mqt/bench/calibration_files/ibm_montreal_calibration.json
burgholzer, line 1: I would rather keep these. Qiskit explicitly adds these to the basis gates. I think one of the reason for that was that, on the pulse level, there is a difference between an idle qubit and a qubit where an identity gate is applied. Something with decoherence or so.
While I believe it is not too important for us at the moment I don't rely see any drawback of keeping the gate in the gateset if IBM does it.

src/mqt/benchviewer/templates/index.html
burgholzer, line 1: Doesn't one of the backends also support the "xx_plus_yy" gate?

burgholzer (approved):
Overall this looks good. Just two minor comments. Feel free to discard them.

src/mqt/benchviewer/templates/index.html
burgholzer: Nevermind 🙃 just saw your commit.

nquetschlich: First I wanted to represent the current state, but since the next release also includes that gate, I added it already. Consequently, we need to re-generate all benchmark files with the next release.

src/mqt/bench/calibration_files/ibm_montreal_calibration.json
nquetschlich: I see. Should be also add it to the others?

burgholzer: Good question. You could maybe cross check with what the official ionq-provider, etc. use as a native gateset.

src/mqt/benchviewer/templates/index.html
burgholzer: That makes sense. Given that we are regenerating quite some circuits it might be worth exploring whether we can get some additions and/or changes in before the next release.

PR #298, opened 2024-02-06
🎨 Improved SupermarQ feature calculation

nquetschlich:
I compared our implementation of the SupermarQ feature calculation with their [implementation](https://github.com/Infleqtion/client-superstaq/blob/main/supermarq-benchmarks/supermarq/converters.py) and copied their code for `liveness` and `program_communication` because it seemed better to me in terms of understadability.

### DIFF
--- a/src/mqt/bench/utils.py
+++ b/src/mqt/bench/utils.py
@@ -15,6 +15,7 @@ import networkx as nx
 import numpy as np
 from pytket import __version__ as __tket_version__
 from qiskit import QuantumCircuit, __qiskit_version__
+from qiskit.converters import circuit_to_dag
 from qiskit_optimization.applications import Maxcut
 
 from mqt.bench.devices import OQCProvider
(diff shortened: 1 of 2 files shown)


src/mqt/bench/utils.py
```diff
@@ -242,50 +243,42 @@ def calc_qubit_index(qargs: list[Qubit], qregs: list[QuantumRegister], index: in
...
+    # Entanglement-ratio = ratio between # of 2-qubit gates and total number of gates in the circuit.
+    entanglement_ratio = len(dag.two_qubit_ops()) / len(dag.gate_nodes())
+
+    # Critical depth = # of 2-qubit gates along the critical path / total # of 2-qubit gates.
+    n_ed = 0
+    for name in {op.name for op in dag.two_qubit_ops()}:
+        try:
+            n_ed += dag.count_ops_longest_path()[name]
+        except KeyError:  # noqa: PERF203
             continue
```
burgholzer, line 279: The computation here can be slightly improved.
First of all, the ruff warning is there for a reason ;o)
Second, this recomputes `count_ops_longest_path()` for every gate name, which is definitely wasteful.
How about the following:
```suggestion
    longest_paths = dag.count_ops_longest_path()
    n_ed = sum([longest_paths[name] for name in {op.name for op in dag.two_qubit_ops()} if name in longest_paths])
```

tests/test_bench.py
```diff
@@ -1059,9 +1059,13 @@ def test_count_qubit_numbers_per_compiler(compiler: str, expected_val: list[int]
...
-    qc = get_benchmark("dj", 1, 5)
-    features = utils.calc_supermarq_features(qc)
-    assert type(features) == utils.SupermarqFeatures
+    qcs = [
+        get_benchmark(benchmark, 1, 5)
+        for benchmark in ["dj", "ghz", "random", "qaoa", "vqe", "qnn", "qpeexact", "qpeinexact", "wstate"]
+    ]
+    for qc in qcs:
+        features = utils.calc_supermarq_features(qc)
+        assert type(features) == utils.SupermarqFeatures
```
burgholzer, line 1068: This is kind of guaranteed by static type checking (mypy) already, so this is kind of a useless assertion.
Can't we properly test something here? For example, that a GHZ state has `parallelism=0`, etc. (preferably in separate tests). For the best tests, we would probably have three scenarios per feature:
- lower extreme (f == 0.)
- higher extreme (f == 1.)
- something in-between (0 < f < 1)

If this is too annoying, then feel free to resolve this and keep the (admittedly suboptimal) tests.

burgholzer (changes requested):
Many thanks for the changes. Looking really good already. I just have one suggestion for the code itself and one for the tests.

burgholzer (approved):
Great. Thanks for all the extra effort. 🎉

PR #299, opened 2024-02-08
MQTPredictor update

flowerthrower:
Mostly bug fixes that became clear when trying to update MQTPredictor based on the new device class.

### DIFF
--- a/src/mqt/bench/devices/rigetti.py
+++ b/src/mqt/bench/devices/rigetti.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-from contextlib import suppress
+import warnings
 from typing import TYPE_CHECKING, TypedDict, cast
 
 if TYPE_CHECKING:
@@ -142,6 +142,14 @@ class RigettiProvider(Provider):
             calibration.t1[qubit] = rigetti_calibration["properties"]["1Q"][str(rigetti_index)]["T1"]
             calibration.t2[qubit] = rigetti_calibration["properties"]["1Q"][str(rigetti_index)]["T2"]
 
+        # To calculate the average value for missing two-qubit gate fidelities
+        cz_lst, cp_lst, xx_plus_yy_lst = [], [], []
+        msg = (
+            "Rigetti device fidelity data is not available for some two-qubit gates."
+            "The average value of the available gates will be used for the missing ones."
+        )
+        warnings.warn(msg, stacklevel=1)
+
         for qubit1, qubit2 in device.coupling_map:
             rigetti_index1 = cls.__to_rigetti_index(qubit1)
             rigetti_index2 = cls.__to_rigetti_index(qubit2)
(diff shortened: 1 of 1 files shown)


src/mqt/bench/devices/rigetti.py
```diff
@@ -94,17 +95,23 @@ def __from_rigetti_index(cls, rigetti_index: int) -> int:
         row = rigetti_index // 100
         column = (rigetti_index % 100) // 10
         ring = rigetti_index % 10
-        return row * (ring_size * columns) + column * ring_size + ring
+        qubit_indx = row * (ring_size * columns) + column * ring_size + ring
+        if is_M3 and qubit_indx >= 70:
```
nquetschlich, line 99: Where does this number 70 come from? Can you maybe a small comment such that others understand why we do this.

src/mqt/bench/devices/rigetti.py
```diff
@@ -94,17 +95,23 @@ def __from_rigetti_index(cls, rigetti_index: int) -> int:
...
+    def __to_rigetti_index(cls, index: int, is_M3: bool) -> int:
         """
         Convert the consecutive index to the Rigetti qubit index.
         Args:
             index: the consecutive index
+            is_M3: if device is Aspen-M3, account for missing qubit 136
 
         Returns: the Rigetti qubit index
         """
+        if is_M3 and index >= 70:
```
nquetschlich, line 113: same here

src/mqt/bench/devices/rigetti.py
```diff
@@ -128,36 +135,69 @@ def import_backend(cls, path: Path) -> Device:
...
             if qubit1 > qubit2:
                 continue
 
             edge = f"{rigetti_index1}-{rigetti_index2}"
             fidelity = {}
-            with suppress(KeyError):
+            try:
                 fidelity["cz"] = rigetti_calibration["properties"]["2Q"][edge]["fCZ"]
-            with suppress(KeyError):
+                cz_lst.append(fidelity["cz"])
```
nquetschlich, line 173: I am not sure I understand why those `_lst` variables are created. Could you briefly explain it to me?

tests/devices/test_rigetti_device_support.py
```diff
@@ -44,6 +44,39 @@ def test_rigetti_aspen_m2_device() -> None:
...
+            device.get_readout_duration(q)
+        for gate in single_qubit_gates:
+            assert isinstance(device.get_single_qubit_gate_fidelity(gate, q), float | int)
+            with pytest.raises(ValueError, match="Single-qubit gate duration values not available."):
+                device.get_single_qubit_gate_duration(gate, q)
+
+    for q0, q1 in device.coupling_map:
+        for gate in two_qubit_gates:
+            assert isinstance(device.get_two_qubit_gate_fidelity(gate, q0, q1), float | int)
+            with pytest.raises(ValueError):  # noqa: PT011
```
nquetschlich, line 78: Why is this ignore necessary?

nquetschlich (changes requested):
Thanks a lot for those changes. I just have a few questions to better understand them.

Additionally, I have one larger comment that probably requires a moderate effort: We should substitute the M3 for the M2 instead of keeping both devices. This means, that all files must be clear of the M2 device including the configuration file. 

For the benchviewer, we currently stick to the M2 device and, therefore, those files must *not* be adjusted. The rationale is, that we will soon update to Qiskit v1.0.0 and then re-create all benchmark files and do not want to do it now additionally since it comes with a significant effort.

src/mqt/bench/devices/rigetti.py
flowerthrower: They make it easy to claculate the average values for the missing ones.
Instead we could store numbers instead and use the update rule like `fid_new += fid_before + ...`.
I think the lists are a bit cleaner bc we would need to introduce a counter for each gate type.

tests/test_bench.py
```diff
@@ -877,21 +876,23 @@ def test_create_benchmarks_from_config(output_path: str) -> None:
...
-    retcode = utils.create_zip_file()
+    zip_path = str(Path("./tests/MQTBench_all.zip").resolve())
+    qasm_path = str(Path("./tests/test_output/").resolve())
+    retcode = utils.create_zip_file(zip_path, qasm_path)
     assert retcode == 0
 
     zip_file = Path(utils.get_zip_file_path())
     assert zip_file.is_file()
-    zip_file.unlink()
+    # zip_file.unlink() TODO: uncomment after update
```
nquetschlich, line 888: What does this `TODO` mean?

tests/test_benchviewer.py
```diff
@@ -282,17 +283,18 @@ def test_prepare_form_input() -> None:
 
 def test_read_mqtbench_all_zip() -> None:
     backend = Backend()
-    with resources.as_file(benchviewer) as benchviewer_path:
-        target_location = str(benchviewer_path / "static/files")
+    with resources.as_file(benchviewer):
+        target_location = str(Path("./tests").resolve())
+        # target_location = str(benchviewer_path / "static/files") TODO: uncomment after update
```
nquetschlich, line 288: see above

tests/test_benchviewer.py
```diff
@@ -282,17 +283,18 @@ def test_prepare_form_input() -> None:
...
 
 
 def test_create_database() -> None:
     backend = Backend()
-
     res_zip = backend.read_mqtbench_all_zip(
         skip_question=True,
-        target_location=str(resources.files("mqt.benchviewer") / "static" / "files"),
+        target_location=str(Path("./tests").resolve()),
+        # target_location=str(resources.files("mqt.benchviewer") / "static" / "files"), TODO: uncomment after update
```
nquetschlich, line 297: see above

tests/test_benchviewer.py
```diff
@@ -313,7 +315,7 @@ def test_create_database() -> None:
 
     res = backend.get_selected_file_paths(input_data)
     assert isinstance(res, list)
-    assert len(res) > 3
+    assert len(res) >= 1  # 3
```
nquetschlich, line 318: All of the comments in this file should be removed it possible.

nquetschlich (changes requested):
As soon as the code comments are clarified and the currently failing test is fixed, this PR is ready to be merged.

PR #310, opened 2024-03-20
Update device calibration

flowerthrower:
Update gate time values to SI units (seconds) and current available calibration data.
This is especially relevant in the light of the new MQTPredictor reward function `expected_success_probability`.


flowerthrower:
I have updated the calibration data to the values specified online (e.g. AWS).
This data comes in various units and is all transformed into SI units when imported using the respective `import_backend` methods.

nquetschlich (approved):
LGTM. Just one small question: we are sure that the units of the other device calibration data sheets are already in seconds?

flowerthrower:
IBM, OQC (and IQM) calibration data is not supplied in seconds (hence transformation during import)
IONQ, Rigetti calibration data is already specified in seconds
Quantinuum - no information on timescales

PR #321, opened 2024-04-19
Add basic CLI around get_benchmark()

ColoredCarrot:
Mainly to be used automatically by tooling.
Options are simple CLI options for now, but should probably become more dynamic in the future (does argparse support dynamic/unregistered options?).

### DIFF
--- a/src/mqt/bench/benchmark_generator.py
+++ b/src/mqt/bench/benchmark_generator.py
@@ -393,8 +393,6 @@ def get_benchmark(
         msg = "compiler_settings must be of type CompilerSettings or None."  # type: ignore[unreachable]
         raise ValueError(msg)
 
-    assert (compiler_settings.tket is not None) or (compiler_settings.qiskit is not None)
-
     independent_level = 1
     if level in ("indep", independent_level):
         if compiler == "qiskit":
--- a/src/mqt/bench/cli.py
+++ b/src/mqt/bench/cli.py
@@ -1,0 +1,61 @@
+import argparse
+
+from mqt.bench import QiskitSettings, TKETSettings, get_benchmark, CompilerSettings
+from pytket import Circuit
+from pytket.qasm import circuit_to_qasm_str
+from qiskit import QuantumCircuit
+from qiskit.qasm2 import dumps as qiskit_circuit_to_str
+
+
+def main() -> None:
+    """
+    Invoke :func:`get_benchmark` exactly once with the specified arguments.
+
+    The resulting QASM string is printed to the standard output stream;
+    no other output is generated.
+    """
+
+    parser = argparse.ArgumentParser(description="Generate a single benchmark")
+    parser.add_argument("--level", type=str,
+                        help='Level to generate benchmarks for ("alg", "indep", "nativegates" or "mapped")',
+                        required=True)
+    parser.add_argument("--algorithm", type=str, help="Name of the benchmark", required=True)
+    parser.add_argument("--num-qubits", type=int, help="Number of Qubits")
+    parser.add_argument("--compiler", type=str, help="Name of the compiler")
+    parser.add_argument("--qiskit-optimization-level", type=int, help="Qiskit compiler optimization level")
+    parser.add_argument("--tket-placement", type=str, help="TKET placement")
+    parser.add_argument("--native-gate-set", type=str, help="Name of the provider")
+    parser.add_argument("--device", type=str, help="Name of the device")
+    args = parser.parse_args()
+
+    qiskit_settings: QiskitSettings | None = None
+    if args.qiskit_optimization_level is not None:
+        qiskit_settings = QiskitSettings(args.qiskit_optimization_level)
+
+    tket_settings: TKETSettings | None = None
+    if args.tket_placement is not None:
+        tket_settings = TKETSettings(args.tket_placement)
+
+    # Note: Assertions about argument validity are in get_benchmark()
+
+    result = get_benchmark(
+        benchmark_name=args.algorithm,
+        level=args.level,
+        circuit_size=args.num_qubits,
+        compiler=args.compiler,
+        compiler_settings=CompilerSettings(
+            qiskit=qiskit_settings,
+            tket=tket_settings,
+        ),
+        provider_name=args.native_gate_set,
+        device_name=args.device,
+    )
+
+    if isinstance(result, QuantumCircuit):
+        qasm_str = qiskit_circuit_to_str(result)
+    elif isinstance(result, Circuit):
+        qasm_str = circuit_to_qasm_str(result)
+    else:
+        raise TypeError(f"Got unknown circuit type from get_benchmark: {type(result)}")
+
+    print(qasm_str)
--- a/src/mqt/bench/cli.py
+++ b/src/mqt/bench/cli.py
@@ -75,7 +75,7 @@ def parse_benchmark_name_and_instance(algorithm: str) -> tuple[str, str | None]:
     as expected by :func:`get_benchmark`.
     """
 
-    if algorithm.startswith("shor_") or algorithm.startswith("groundstate_"):
+    if algorithm.startswith(("shor_", "groundstate_")):
         as_list = algorithm.split("_", 2)
         assert len(as_list) == 2
         return cast(tuple[str, str], tuple(as_list))
--- a/src/mqt/bench/cli.py
+++ b/src/mqt/bench/cli.py
@@ -25,7 +25,7 @@ def main() -> None:
         required=True,
     )
     parser.add_argument("--algorithm", type=str, help="Name of the benchmark", required=True)
-    parser.add_argument("--num-qubits", type=int, help="Number of Qubits")
+    parser.add_argument("--num-qubits", type=int, help="Number of Qubits", required=True)
     parser.add_argument("--compiler", type=str, help="Name of the compiler")
     parser.add_argument("--qiskit-optimization-level", type=int, help="Qiskit compiler optimization level")
     parser.add_argument("--tket-placement", type=str, help="TKET placement")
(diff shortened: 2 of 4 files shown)


src/mqt/bench/benchmark_generator.py
```diff
@@ -393,8 +393,6 @@ def get_benchmark(
         msg = "compiler_settings must be of type CompilerSettings or None."  # type: ignore[unreachable]
         raise ValueError(msg)
 
-    assert (compiler_settings.tket is not None) or (compiler_settings.qiskit is not None)
-
```
burgholzer, line 397: Any reason why this assertion has been removed? If it's because of how the CLI sets both to `None` if none of the parameters are passed to the CLI, the suggestion from that file should fix that condition.
That should allow you to add back the assertion here.

src/mqt/bench/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    parser = argparse.ArgumentParser(description="Generate a single benchmark")
+    parser.add_argument(
+        "--level",
+        type=str,
+        help='Level to generate benchmarks for ("alg", "indep", "nativegates" or "mapped")',
+        required=True,
+    )
+    parser.add_argument("--algorithm", type=str, help="Name of the benchmark", required=True)
```
burgholzer, line 28: I'd guess that this argument should also be required. Doesn't really make sense to generate a benchmark without knowing which algorithm to generate.

src/mqt/bench/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    parser.add_argument("--num-qubits", type=int, help="Number of Qubits")
```
burgholzer, line 29: In the same spirit as above, this probably also has to be mandatory. Otherwise, it is not clear which benchmark to generate.
Note that there are certain non-scalable benchmarks that rather take a `str` that indicates the instance. You can see the special handling here: https://github.com/cda-tum/mqt-bench/blob/60e1fa7dad1584c3497384a9a906d64acc7f14a9/src/mqt/bench/benchmark_generator.py#L348-L379
I could see multiple different ways of handling that. Either is fine for me.

src/mqt/bench/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    parser.add_argument("--compiler", type=str, help="Name of the compiler")
+    parser.add_argument("--qiskit-optimization-level", type=int, help="Qiskit compiler optimization level")
+    parser.add_argument("--tket-placement", type=str, help="TKET placement")
+    parser.add_argument("--native-gate-set", type=str, help="Name of the provider")
+    parser.add_argument("--device", type=str, help="Name of the device")
+    args = parser.parse_args()
+
+    qiskit_settings: QiskitSettings | None = None
```
burgholzer, line 37:
```suggestion
    qiskit_settings = QiskitSettings()
```
This sets the default values properly.

src/mqt/bench/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    if args.qiskit_optimization_level is not None:
+        qiskit_settings = QiskitSettings(args.qiskit_optimization_level)
+
+    tket_settings: TKETSettings | None = None
```
burgholzer, line 41:
```suggestion
    tket_settings = TKETSettings()
```
This sets the default values properly.

src/mqt/bench/cli.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if isinstance(result, QuantumCircuit):
+        qasm_str = qiskit_circuit_to_str(result)
+    elif isinstance(result, Circuit):
+        qasm_str = circuit_to_qasm_str(result)
+    else:
+        msg = f"Got unknown circuit type from get_benchmark: {type(result)}"
+        raise TypeError(msg)
+
+    print(qasm_str)
```
burgholzer, line 68:
```suggestion
    if isinstance(result, QuantumCircuit):
        print(qiskit_circuit_to_str(result))
        return
    
    print(circuit_to_qasm_str(result))
```
The else branch above was unreachable per definition (i.e., by static typing), so the code here can be simplified a little.

burgholzer (changes requested):
Hey 👋🏼 
Many thanks for this first contribution! As discussed in other channels, this seems like a useful feature and the current PR is a solid basis for that.
I added some inline suggestions on some small improvements and some todos to make this feature complete. Those shouldn't really take long to incorporate.

(If you are not aware of that: if you go to the "Files changed" tab on GitHub, you can batch up suggestions into a single commit and apply them in one go directly from the web. This also resolves the related comments)

As already hinted at by the CLI, this PR is missing one more thing in addition to these small comments: test (coverage). There is a convenient way to test CLIs with unittests:
- you add the `pytest-console-scripts` dependency as illustrated here: https://github.com/cda-tum/mqt-core/blob/4d6f788df95513b9c14db938bae69f1683e26f3d/pyproject.toml#L39
- you write tests similar to https://github.com/cda-tum/mqt-core/blob/main/test/python/test_cli.py

Those tests need not be massive, but they should cover the most common scenarios in which a user would run the CI.

Once those changes are implemented, this should be good to go and a really nice addition to the library. Many thanks 🙏🏼

src/mqt/bench/benchmark_generator.py
ColoredCarrot: It's not directly related to the CLI. IMO, the assertion at this place is "wrong" since it triggers even when requesting the `alg` level, which isn't associated with a compiler.
Additionally, the assertion is essentially duplicated by more specific assertions later on (which check that the settings corresponding to the requested compiler are set). In my view, that obviates this assertion (DRY and all).

src/mqt/bench/cli.py
ColoredCarrot: It is `required` though, isn't it? :D

src/mqt/bench/cli.py
ColoredCarrot: Making this mandatory makes sense to me.
The benchmark instance thing is interesting. How about splitting those benchmarks up, at least from the user's POV, and removing the concept of instances from the interface? Should be feasible since neither shor nor groundstate parse the string in any way, and just have a fixed set of options.

src/mqt/bench/benchmark_generator.py
burgholzer: good points! got me convinced ;-)

src/mqt/bench/cli.py
burgholzer: Yap. Nevermind. Must have overlooked that while reviewing on the phone 🙃

src/mqt/bench/cli.py
burgholzer: Just so that I get it right what you are proposing (which I think makes sense):
instead of having to specify `benchmark_name="shor"` and `benchmark_instance_name="xlarge"`, you would rather make this accessible as `benchmark_name="shor_xlarge"` for example?
Did I get that right? (in that case, these qubits wouldn't need a qubit number then)

Another special case in the benchmarks right now is that some benchmarks (grover and qwalk) have an additional parameter (how they handle ancillary qubits). In the UI, these are currently distinguished as `Grover's (no ancilla)` and `Grover's (v-chain)`. 
Any solution that solves the above, would also be applicable here.

ColoredCarrot: Yes, that's what I'm proposing (with nice display names in the web UI that make this obvious).

The ancilla handling seems to already work that way: `get_benchmark` does some special parsing of the given `benchmark_name` and then invokes `create_circuit` with the extra parameter.
I would prefer to completely get rid of this special case at this point, e.g.: Instead of loading python files by benchmark names, which requires one file per benchmark, have a `dict[benchmark_name, BenchmarkFactory]` (where `BenchmarkFactory` is akin to `(bench_name) -> QuantumCircuit`). Then, the same factory can be registered for multiple bench names, and the factory itself can be responsible for parsing the name.

ColoredCarrot:
Re. tests: `dumps(get_benchmark(...))` seems to yield non-deterministic results. I'm lacking the domain knowledge here; is there some easy way to verify that the printed QASM string corresponds to specific options?
If not, I'll just do some basic sanity checks (like *these inputs should work and the result should contain "OPENQASM 2.0;"* + *these inputs shouldn't work*).

src/mqt/bench/cli.py
burgholzer: That sounds like a very reasonable change that would improve the library structure.
Might be a slightly bigger change though. How about "just making it work" here and doing the proposed refactor in a separate PR?

ColoredCarrot: Yes, absolutely agree 😃

burgholzer:
> Re. tests: `dumps(get_benchmark(...))` seems to yield non-deterministic results. I'm lacking the domain knowledge here; is there some easy way to verify that the printed QASM string corresponds to specific options?
> 
> If not, I'll just do some basic sanity checks (like *these inputs should work and the result should contain "OPENQASM 2.0;"* + *these inputs shouldn't work*).

Yeah. Some of the benchmarks involve steps that aren't very easy to seed for reproducibility. Thus, sometimes a check like comparing the result of the CLI with the result of the same `get_benchmark` call might fail.
However there should be some
Benchmarks that are fairly deterministic. Just at the top of my head:
- the "ghz" benchmark should work deterministically under almost all settings
- "shor" as a non-scalable example should be deterministic as well
- given that the benchmarks with ancillary settings are handled similar to regular instances, there should be no need to test them here.

As stated in the comment at the top, the tests here really don't need to be too exhaustive. If the above helps to create a working set of tests, that should be fine 😌

burgholzer (approved):
Many thanks for the additional changes and the quick workaround.
Everything looks good to me now!
Would really appreciate another PR resolving the special casing as discussed in the comment above!

PR #335, opened 2024-05-24
🧱 switched to different CI workflow source

nquetschlich:


burgholzer (approved):
Just one clarification question here. Otherwise LGTM

PR #353, opened 2024-07-07
📌 removed dep upper cap

nquetschlich:


burgholzer (approved):
LGTM 👍🏻 let's keep evaluating on how much this saves us in maintenance overhead or if it just causes more work 😌

PR #367, opened 2024-08-20
👷 update framework

nquetschlich:
This PR
- updates the pre-commit config
- updates the CI/CD workflows
- adds python 3.13 support

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   python-packaging:
     name: 🐍 Packaging
-    uses: cda-tum/mqt-workflows/.github/workflows/reusable-python-packaging.yml@v1.1.5
+    uses: cda-tum/mqt-workflows/.github/workflows/reusable-python-packaging.yml@v1.3
     with:
       pure-python: true
 
@@ -20,9 +20,11 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      url: https://pypi.org/p/mqt.bench
+      url: https://pypi.org/p/mqt.predictor
     permissions:
       id-token: write
+      attestations: write
+      contents: read
     needs: [python-packaging]
     steps:
       - uses: actions/download-artifact@v4
@@ -30,6 +32,10 @@ jobs:
           pattern: cibw-*
           path: dist
           merge-multiple: true
+      - name: Generate artifact attestation for sdist and wheel(s)
+        uses: actions/attest-build-provenance@v1.4.0
+        with:
+          subject-path: "dist/*"
       - uses: pypa/gh-action-pypi-publish@release/v1
 
   upload_webserver:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -113,7 +113,7 @@ repos:
 
   # Check best practices for scientific Python code
   - repo: https://github.com/scientific-python/cookie
-    rev: 2024.04.23
+    rev: 2024.08.19
     hooks:
       - id: sp-repo-review
         additional_dependencies: ["repo-review[cli]"]
@@ -128,6 +128,6 @@ repos:
 
   # Check the pyproject.toml file
   - repo: https://github.com/henryiii/validate-pyproject-schema-store
-    rev: 2024.08.08
+    rev: 2024.08.19
     hooks:
       - id: validate-pyproject
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,9 +60,7 @@ repos:
     hooks:
       - id: ruff
         args: ["--fix", "--show-fixes"]
-        types_or: [python, pyi, jupyter]
       - id: ruff-format
-        types_or: [python, pyi, jupyter]
 
   # Also run Black on examples in the documentation
   - repo: https://github.com/adamchainz/blacken-docs
@@ -102,7 +100,7 @@ repos:
     rev: v1.23.6
     hooks:
       - id: typos
-        exclude: .html
+        exclude: legal.html
 
   # Catch common capitalization mistakes
   - repo: local
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -109,7 +109,7 @@ repos:
         name: Disallow improper capitalization
         language: pygrep
         entry: PyBind|Numpy|Cmake|CCache|Github|PyTest|Mqt|Tum
-        exclude: \.pre-commit-config\.yaml|\.joblib$
+        exclude: .pre-commit-config.yaml
 
   # Check best practices for scientific Python code
   - repo: https://github.com/scientific-python/cookie
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ ci:
 repos:
   # Standard hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: "v4.6.0"
+    rev: v4.6.0
     hooks:
       - id: check-added-large-files
       - id: check-case-conflict
@@ -28,9 +28,20 @@ repos:
       - id: mixed-line-ending
       - id: trailing-whitespace
 
+  # Clean jupyter notebooks
+  - repo: https://github.com/srstevenson/nb-clean
+    rev: 3.3.0
+    hooks:
+      - id: nb-clean
+        args:
+          - --remove-empty-cells
+          - --preserve-cell-metadata
+          - raw_mimetype
+          - --
+
   # Handling unwanted unicode characters
   - repo: https://github.com/sirosen/texthooks
-    rev: "0.6.6"
+    rev: 0.6.7
     hooks:
       - id: fix-ligatures
       - id: fix-smartquotes
@@ -43,28 +54,9 @@ repos:
       - id: rst-directive-colons
       - id: rst-inline-touching-normal
 
-  # Clean jupyter notebooks
-  - repo: https://github.com/srstevenson/nb-clean
-    rev: "3.3.0"
-    hooks:
-      - id: nb-clean
-  # Check for spelling
-  - repo: https://github.com/codespell-project/codespell
-    rev: "v2.3.0"
-    hooks:
-      - id: codespell
-        args: ["-L", "wille,linz,fro"]
-        exclude: "mqt/bench/viewer/templates/legal.html"
-
-  # Format configuration files with prettier
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
-    hooks:
-      - id: prettier
-        types_or: [yaml, markdown, html, css, javascript, json]
-
+  # Python linting and formatting using ruff
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.5
+    rev: v0.6.1
     hooks:
       - id: ruff
         args: ["--fix", "--show-fixes"]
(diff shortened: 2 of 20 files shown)


nquetschlich:
The currently thrown errors `'StatePreparation' object has no attribute '_gates_to_uncompute'` are apparently a bug that will soon be fixed in qiskit, see https://github.com/Qiskit/qiskit/issues/12969 for details.

burgholzer:
> The currently thrown errors `'StatePreparation' object has no attribute '_gates_to_uncompute'` are apparently a bug that will soon be fixed in qiskit, see https://github.com/Qiskit/qiskit/issues/12969 for details.

Alright. Given that this is supposedly fixed in 1.2.1, the easiest fix would be to exclude the 1.2.0 release here, I.e.
```
"qiskit!=1.2.0", # skip faulty qiskit release
```

.github/workflows/cd.yml
```diff
@@ -20,16 +20,22 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      url: https://pypi.org/p/mqt.bench
+      url: https://pypi.org/p/mqt.predictor
```
burgholzer, line 23:
```suggestion
      url: https://pypi.org/p/mqt.bench
```

.github/workflows/cd.yml
```diff
@@ -20,16 +20,22 @@ jobs:
...
+      contents: read
     needs: [python-packaging]
     steps:
       - uses: actions/download-artifact@v4
         with:
           pattern: cibw-*
           path: dist
           merge-multiple: true
+      - name: Generate artifact attestation for sdist and wheel(s)
+        uses: actions/attest-build-provenance@v1.4.0
```
burgholzer, line 36:
```suggestion
        uses: actions/attest-build-provenance@v1.4.1
```
IIRC, there was an update here lately. Doesn't hurt to cover this here directly.

.pre-commit-config.yaml
```diff
@@ -43,28 +54,9 @@ repos:
...
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
-    hooks:
-      - id: prettier
-        types_or: [yaml, markdown, html, css, javascript, json]
-
+  # Python linting and formatting using ruff
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.5
+    rev: v0.6.1
```
burgholzer, line 59: Unfortunately can't directly comment on the respective line, but the `types_or` lines in both ruff pre-commit checks down below can now be removed since ruff 0.6 automatically runs on Jupyter notebooks and is no longer opt-in.

.pre-commit-config.yaml
```diff
@@ -97,18 +97,39 @@ repos:
           - types-tqdm
           - types-flask
 
+  # Check for spelling
+  - repo: https://github.com/crate-ci/typos
+    rev: v1.23.6
+    hooks:
+      - id: typos
+        exclude: .html
```
burgholzer, line 105: I am not 100% sure this is the best way to go about this. This would hide all potential typos in HTML pages.
I think I get why you put this here because I guess it autocorrects some html elements, but I think the configuration options of the `typos` project are flexible enough to exclude these separately.

.pre-commit-config.yaml
```diff
@@ -97,18 +97,39 @@ repos:
...
+
   # Catch common capitalization mistakes
   - repo: local
     hooks:
       - id: disallow-caps
         name: Disallow improper capitalization
         language: pygrep
         entry: PyBind|Numpy|Cmake|CCache|Github|PyTest|Mqt|Tum
-        exclude: .pre-commit-config.yaml
+        exclude: \.pre-commit-config\.yaml|\.joblib$
```
burgholzer, line 114: Two questions on this one:
- Why the sudden escapes of the dots, and
- Why the additional `.joblib` exclude that was not necessary before

.pre-commit-config.yaml
```diff
@@ -97,18 +97,39 @@ repos:
...
+    hooks:
+      - id: check-dependabot
+      - id: check-github-workflows
+      - id: check-readthedocs
+
+  # Check the pyproject.toml file
+  - repo: https://github.com/henryiii/validate-pyproject-schema-store
+    rev: 2024.08.08
+    hooks:
+      - id: validate-pyproject
```
burgholzer, line 135: Might be worth running `pre-commit autoupdate` once because I think that new versions of `cookie` and the `validate-pyproject-schema-store` came out in the last couple of days.
I believe these also make the `cookie` ignores in the `pyproject.toml` obsolete.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dynamic = ["version"]
 dependencies = [
     "pytket-qiskit>=0.53.0",  #  manages the dependencies for qiskit and tket in a combined fashion
     "pytket>=1.29.0",  # lowest version that supports the used pytket AutoRebase pass instead of auto_rebase
+    "qiskit!=1.2.0",  # there is a bug in 1.20.0 that causes an error some benchmarks, see https://github.com/Qiskit/qiskit/issues/12969
```
burgholzer, line 24:
```suggestion
    "qiskit!=1.2.0",  # there is a bug in 1.2.0 that causes an error some benchmarks, see https://github.com/Qiskit/qiskit/issues/12969
```

pyproject.toml
```diff
@@ -92,15 +94,29 @@ log_cli_level = "INFO"
...
+    'ignore:.*The class ``qiskit.providers.models.backendconfiguration.BackendConfiguration`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.providers.models.backendconfiguration.QasmBackendConfiguration`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.providers.models.backendproperties.BackendProperties`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.qobj.pulse_qobj.PulseLibraryItem`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.primitives.base.base_estimator.BaseEstimator`` is deprecated as of qiskit 1.2*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.primitives.base.base_sampler.BaseSampler`` is deprecated as of qiskit 1.2*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.primitives.sampler.Sampler`` is deprecated as of qiskit 1.2*:DeprecationWarning:mqt.*',
+    'ignore:.*The class ``qiskit.primitives.estimator.Estimator`` is deprecated as of qiskit 1.2*:DeprecationWarning:mqt.*',
```
burgholzer, line 116: If those are coming from our own library, then we should fix them. There are direct replacements for the sampler and the estimator.

pyproject.toml
```diff
@@ -92,15 +94,29 @@ log_cli_level = "INFO"
...
+    'ignore:.*invalid value encountered in det.*:RuntimeWarning:',
+    'ignore::DeprecationWarning:.*(docplex).*',
+    'ignore::SyntaxWarning:.*(docplex).*',
+    'ignore:.*The abstract Provider and ProviderV1 classes are deprecated.*:DeprecationWarning:qiskit.*',
+    'ignore:.*qiskit.providers.models is deprecated since Qiskit 1.2*:DeprecationWarning:',
+    'ignore:.*Treating CircuitInstruction as an iterable is deprecated legacy behavior since Qiskit 1.2*:DeprecationWarning:pytket.*',
+    'ignore:.*The class ``qiskit.providers.models.backendconfiguration.GateConfig`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.providers.models.backendconfiguration.BackendConfiguration`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.providers.models.backendconfiguration.QasmBackendConfiguration`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
+    'ignore:.*The class ``qiskit.providers.models.backendproperties.BackendProperties`` is deprecated as of qiskit 1.2.*:DeprecationWarning:qiskit.*',
```
burgholzer, line 111:
```suggestion
    'ignore:.*qiskit.providers.models.* is deprecated since Qiskit 1.2*:DeprecationWarning:',
```
Fairly sure that something like the suggestion should work to unify all these ignores.
Something similar can definitely be done for the `qiskit.primitives.*` warnings.

pyproject.toml
```diff
@@ -202,3 +218,28 @@ isort.required-imports = ["from __future__ import annotations"]
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.cibuildwheel]
+# Enable free-threaded support
+free-threaded-support = true
```
burgholzer, line 224:
```suggestion
```
MQT Bench does not use cibuildwheel, so this setting is obsolete.

pyproject.toml
```diff
@@ -202,3 +218,28 @@ isort.required-imports = ["from __future__ import annotations"]
...
+
+[tool.repo-review]
+ignore = [
+    "PC160", # "Uses codespell" -> switched to https://github.com/crate-ci/typos
+    "PC180", # "Uses prettier" -> switched to different prettier-mirror that is not recognized by the check
+    "RF003", # "src directory specified if used" -> obsolete for new ruff versions
+]
```
burgholzer, line 231: As stated above, these might become obsolete after a `pre-commit autoupdate`

src/mqt/bench/benchmarks/random.py
```diff
@@ -1,4 +1,4 @@
-"""Random benchmark definition."""
+"""Random benchmark definition."""  # noqa: A005
```
burgholzer, line 1: I'd rather prefer to rename this file to something more suitable like `random_circuit` instead of ignoring the warning.

src/mqt/bench/devices/ibm.py
burgholzer, line 1: Given that we only support importing BackendsV2 here now, the code at the end of the file can be simplified a little bit. The type check is handled by mypy already.
The body of the `import_qiskit_backend` method can be directly replaced with the body of the `__import_backend_v2` method, which can then be deleted.

burgholzer (changes requested):
Thanks for your work on this. Just did a very quick review and left some comments. Hopefully shouldn't take too long to incorporate. Similar feedback probably also applies to the predictor PR.

pyproject.toml
burgholzer, line 1: Can't comment on the line, but you can also remove the `src = [ "src"]` from the ruff config

.pre-commit-config.yaml
nquetschlich: Done.

.pre-commit-config.yaml
nquetschlich: I now only exclude the (German) `legal.html`.

.pre-commit-config.yaml
nquetschlich: I reverted these changes.

.pre-commit-config.yaml
nquetschlich: Done and removed the ignores.

pyproject.toml
nquetschlich: Done.

pyproject.toml
nquetschlich: Done, but did it slightly differently and therefore did not commit the suggestion itself.

pyproject.toml
nquetschlich: Done.

src/mqt/bench/benchmarks/random.py
nquetschlich: I thought about that but was unsure because we would need to re-generate MQTBench since also the name of the benchmark when calling `get_benchmark` would change.
Similarly, we would need to update `mqt.bench.viewer` including the backend logic which would either lead to a major version update because of breaking changes or multiple checks at different places to avoid the breaking change and make both names work. 

Do you see any other way? To me it feels like this issue should be treated whenever we want to re-generate MQTBench (including a major version update) and for now I would be in favor of the ignore.

burgholzer: I see the point. Not worth the effort here.
Let's do this if we start moving away from qiskit algorithms as a whole, which will probably be a breaking change in any case.

burgholzer (approved):
Thanks for the additional changes! LGTM now ✅

PR #394, opened 2024-10-02
🐛 Verify provider and device availability based on selected level

Drewniok:
This PR addresses a bug where the provider's availability was checked, even when the level was set to ``mapped``. This caused an issue if ``provider_name`` was not provided in ``mqt.bench.cli`` while using the ``mapped`` level.

### DIFF
--- a/src/mqt/bench/benchmark_generator.py
+++ b/src/mqt/bench/benchmark_generator.py
@@ -439,11 +439,12 @@ def get_benchmark(
         if compiler == "tket":
             return tket_helper.get_native_gates_level(qc, provider, circuit_size, False, True)
 
+    if device_name not in get_available_device_names():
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
Thanks for catching this. Seems like a simple fix 👍🏻

Drewniok:
@burgholzer Do you know why there seems to be a problem with the token?

burgholzer:
> @burgholzer Do you know why there seems to be a problem with the token? 

Yeah. Pretty sure we are hitting: https://github.com/actions/toolkit/issues/952
That is, because this PR is from a fork and we do not explicitly set the right permissions (`id-token: write`), the request fails.
We might be able to fix this by explicitly setting the permissions.

For this PR, I'll just go ahead and admin merge 😌

PR #404, opened 2024-10-16
🔥Remove most of the qiskit application module benchmarks

### LINKED ISSUE #390: ✨ Removal of deprecated dependencies

There are several benchmarks that are based on the qiskit application modules. However, these modules are not officially supported by IBM and, therefore, might be a dependency users do not want to have. 

The benchmarks based on these modules should be adjusted and in the worst case removed if no other solution is feasible.

burgholzer:
This PR removes most of the qiskit application module benchmarks. The corresponding packages are hardly maintained and the corresponding benchmarks where simple ansatz functions that are available as part of MQT Bench anyways.
Based on the fact that hardly any optimization is being conducted as part of the generation process, these application benchmarks weren't that different from the basic ansatz function benchmarks.

This reduces the number of available benchmarks. At the same time, it considerably reduces the dependencies of MQT Bench.
Overall, this still marks a major breaking change in the benchmarks offered by MQT Bench.

Fixes #390


nquetschlich:
this will be implemented as part of #429

PR #429, opened 2024-11-20
🎨 MQT Bench Update: New device structure and device, code refactoring, and further improvements

### LINKED ISSUE #220: Add 433-Qubit IBM Device with new native gate set
As soon as a `qiskit.providers.fake_provider` is provided for the new 433-qubit `ibmq_seattle` device is available, it should be integrated into MQT Bench.

Since that device has now a different gate set (with `ECR` as its entangling gate), those changes must also be reflected in the overall UI of MQT Bench.

### LINKED ISSUE #368: Rename `random` benchmark because of shadowing python's module with the same name
I'd rather prefer to rename this file to something more suitable like `random_circuit` instead of ignoring the warning.

_Originally posted by @burgholzer in https://github.com/cda-tum/mqt-bench/pull/367#discussion_r1726630633_

### LINKED ISSUE #387: Remove the `Provider` concept and replace it with base classes of `Device`s
One possible idea in that regard would be to kind of refactor the device handling to remove the `Provider` concept and replace it with base classes of `Device`s that implement shared functionality.

_Originally posted by @burgholzer in https://github.com/cda-tum/mqt-bench/pull/381#discussion_r1767656527_

### LINKED ISSUE #388: ✨ Improving documentation to make it more explicit which benchmarks trigger an optimization procedure when generated

Some of the benchmarks (e.g., VQE, QAOA, Portfolio benchmarks, etc.) have parameterized gates. To determine their values, a short optimization procedure is triggered which still takes some time. 

The documentation should highlight which benchmarks are affected by this. Alternatively, one could think about assigning random parameter values since the optimized ones are most likely also far from optimal.

### LINKED ISSUE #389: ✨ OpenQASM 3.0 support

Currently, all pre-generated benchmarks are in the qasm2 format. This is problematic, because it cannot represent arbitrary quantum gates and especially not parameterized ones. However, it would be desirable to also include benchmarks with symbolic parameters instead of specific assigned parameter values.

Support of OpenQASM 3.0 to mitigate this problem.

### LINKED ISSUE #390: ✨ Removal of deprecated dependencies

nquetschlich:
This PR 
1) removes the `Provider` class (resolves #387)
2) removes the `placement` option for the mapped TKET quantum circuits
3) removes the frontend and evaluation submodules
4) adds a new IBM device: `ibmq_torino` (and, with that, resolves #220 kind of)
5) improves the `Device` class structure
6) refactors the tests to accelerate the test execution time
7) renames the `random` circuit (resolves #368) 
8) adds a qasm3 export function (resolves #389) 
9) adds the `Bernstein-Vazarani` algorithm
10) removes all benchmarks relying on not-maintained-anymore dependencies `qiskit_algorithms`, `qiskit_optimization`, and `qiskit_nature` (making #404 obsolete and resolves #390 as well as resolves #388)
11) renames the vqe ansatz circuit benchmarks and rewrote the qaoa benchmark to not use the qiskit application libraries anymore
12) adds a algorithm level benchmark generation function for qiskit that only supports qasm3

As a consequence of this PR, the `get_benchmark` function parameter name changed since it does not need the provider anymore but a gateset. However, this allows one to easily generate benchmarks for custom gatesets.

This PR comes with multiple breaking changes.


burgholzer:
I just rebased this branch on the current main and cleaned up the history a little bit.
Hopefully this makes it easier to extract individual PRs out of here.
I'd hope to iteratively move something out of the PR here into a separate PR and then rebase here.
Let's see how that goes.

nquetschlich:
Due to the size of this PR and the many different topics it touches, it was split into multiple PRs that are merged by now:
- #549 resolved mentioned issues 1, 4, 5, 6
- #510 resolved issue 2
- #504 resolved issue 3
- #508 resolved issues 7 and 11
- #518 resolved issues 8 and 12
- #505 resolved issue 9
- #404 resolves issue 10

Therefore, this PR is closed.

PR #463, opened 2025-01-13
Add Two Benchmarking Quantum Circuits: quarkcardinality and quarkcopula

fkiwit:
This pull request adds two benchmarking quantum circuits from the generative modeling application of the QUARK framework:

 - quarkcardinality.py
 - quarkcopula.py

Both circuits are part of the QUARK framework: [QUARK GitHub Repository](https://github.com/QUARK-framework/QUARK)


nquetschlich:
Thanks a lot for contributing your benchmarks to MQT Bench @fkiwit. As discussed, we will integrate them with the update of MQT Bench that allows for OpenQASM 3 export. I just leave a few todos we have to complete when integrating the benchmarks into the updated MQT Bench version (this is more a reminder for me rather than it is meant that you need to do these things):
- replace random parameter values with actual parameters (and maybe an option to assign random ones)
- add these benchmarks to the config file (if still present)
- add documentation for both circuits for the website and read the docs

nquetschlich:
implemented in #541

PR #505, opened 2025-04-10
✨ Added Bernstein-Vazirani Algorithm

simon1hofmann:
Extracted from #429:
9. adds the Bernstein-Vazarani algorithm


src/mqt/bench/benchmarks/bv.py
```diff
@@ -0,0 +1,73 @@
...
+from qiskit import QuantumCircuit
+
+
+def create_circuit(num_qubits: int, dynamic: bool = False, hidden_string: str | None = None) -> QuantumCircuit:
+    """Create a quantum circuit for the Bernstein-Vazirani algorithm.
+
+    Arguments:
+        num_qubits: Total number of qubits in the circuit (including the flag qubit).
+        hidden_string: The hidden bitstring to be found (default: alternating pattern of 1 and 0).
+        dynamic: Whether to use a dynamic layout (default: False).
```
nquetschlich, line 14: The order of the arguments does not correspond to the parameter order in the function definition.

nquetschlich (changes requested):
In general this looks good to me. Ideally, we would already add some basic test functionality here and then only check in the end after all PRs have merged for uncovered code.

nquetschlich (approved):

PR #506, opened 2025-04-10
🎨 Re-implemented AE without Qiskit Application Modules

simon1hofmann:
Extracted from https://github.com/cda-tum/mqt-bench/pull/429


src/mqt/bench/benchmarks/ae.py
```diff
@@ -4,66 +4,66 @@
...
 
-def get_estimation_problem() -> EstimationProblem:
-    """Returns a estimation problem instance for a fixed p value."""
-    p = 0.2
 
-    """A circuit representing the Bernoulli A operator."""
-    a = QuantumCircuit(1)
-    theta_p = 2 * np.arcsin(np.sqrt(p))
-    a.ry(theta_p, 0)
+def inverse_qft(num_qubits: int) -> QuantumCircuit:
```
nquetschlich, line 61: This looks like a dead function that can be removed.

src/mqt/bench/benchmarks/ae.py
```diff
@@ -4,66 +4,66 @@
...
+    a.ry(theta_p, 0)
 
-    def __init__(self, probability: float) -> None:
-        """Initialize the Bernoulli Q operator."""
-        super().__init__(1)  # circuit on 1 qubit
+    # Apply Hadamard gates to the evaluation qubits
+    qc.h(range(num_eval_qubits))
 
-        self._theta_p = 2 * np.arcsin(np.sqrt(probability))
-        self.ry(2 * self._theta_p, 0)
```
nquetschlich, line 36: Here, there is a factor of 2 in there and that one is removed in the update. Could you briefly check whether this makes sense?

nquetschlich (changes requested):
I added two comments.

nquetschlich (approved):

PR #507, opened 2025-04-10
🔥 Removed all Benchmarks relying on `qiskit_optimization`, `qiskit_nature`, and `qiskit_algorithm` Dependencies

simon1hofmann:
Extracted from https://github.com/cda-tum/mqt-bench/pull/429:
10. removes all benchmarks relying on not-maintained-anymore dependencies qiskit_algorithms, qiskit_optimization, and qiskit_nature (making https://github.com/cda-tum/mqt-bench/pull/404 obsolete and resolves https://github.com/cda-tum/mqt-bench/issues/390 as well as resolves https://github.com/cda-tum/mqt-bench/issues/388)


src/mqt/bench/utils.py
```diff
@@ -55,14 +51,8 @@ def get_supported_benchmarks() -> list[str]:
         "realamprandom",
```
nquetschlich, line 51: I suggest to rename all VQE ansatz circuits (realamprandom, su2random, and twolocalrandom) with "vqe-<former name>" to make clear, that we still support VQE circuits. Otherwise, people could interpret it differently.

nquetschlich (approved):

src/mqt/bench/utils.py
simon1hofmann: Isn't that done in #508?

nquetschlich: Yes, I forgot about that and saw that PR later.

PR #509, opened 2025-04-10
🎨 Miscellaneous Fixes

simon1hofmann:
Extracted from #429


tests/test_bench.py
```diff
@@ -314,22 +314,6 @@ def test_routing() -> None:
     assert qc.depth() > 0
 
 
-def test_unidirectional_coupling_map() -> None:
```
nquetschlich, line 317: I think we should keep this test.

nquetschlich (changes requested):

nquetschlich (approved):

PR #518, opened 2025-04-14
✨ Added QASM3 support

simon1hofmann:
Extracted from #429:
8. adds a qasm3 export function (resolves https://github.com/cda-tum/mqt-bench/issues/389)
12. adds a algorithm level benchmark generation function for qiskit that only supports qasm3


burgholzer (approved):
Many thanks for adding this 🙏🏼 this help is much appreciated.
I just pushed two small fix ups; one for updating the lockfile and one for making the decomposition slighlty more specific. Technically, this is a bug in Qiskit that could be worth reporting. Or at least worth checking, if this is fixed in more recent versions.

The PR itself looks good to me as it is given the current (changing) state of the code base. So let's get this in 🚀 
However, looking through the code triggered a couple of follow-up request. Some of these might not be fully related to this PR specifically, but they are more general observations. Maybe @nquetschlich or @simon1hofmann could turn these into issues to follow up on after this PR is completed.

- Based on offline discussion, dropping support for TKET seems reasonable and simplifies quite some of the code base. Also with regard to exporting. It opens up the door for directly supporting Qiskit 2.0 as well as Python 3.9.
- So, after dropping TKET and related dependencies, Python 3.9 support should be added back.
- Similarly, compatibility with Qiskit 2.0 should be tested and deprecation warnings should be resolved.
- We should generalize the output logic from purely OpenQASM (2 or 3) to more general output formats. Every function parameter in this PR that is called "qasm_format" should most likely be called "output_format" and should accept values from a string enum. Once a qiskit `QuantumCircuit` is the only immediate representation we are relying on, we can really streamline the output routines and move them to a dedicated module (no more `utils` 🙃). That module should, for the beginning, definitely contain routines for exporting to QASM2, QASM3, and Qiskit's QPY, which should be easy enough to support out of the box. In the future, I would imagine this to be extended by plugins, e.g., for `tket`, `catalyst`, `cuda-q`, `mqt-core-ir`, etc. The whole point is to decouple the generation logic from the output generation logic.
- Once TKET is removed, there is no need anymore to use a `qiskit_helpers` file. The code can be reorganized to simply rely on qiskit instead of calling out to the helpers.
- Given how it is completely unrealistic to "generate all benchmarks" the respective methods can, in my opinion, be removed from the benchmark_generator. That removes another bunch of code.
- MQT Bench currently does not use PEP 735 dependency groups. Inspiration for how to add these may come from MQT QECC, which recently switched to a pure-python package and has the most modern setup in that regard (see https://github.com/cda-tum/mqt-qecc/pull/390)
- MQT Bench still uses setuptools as a build backend. There are better and more modern alternatives that also break less often. The concrete proposal here would be to switch to hatchling as a build backend. This is also covered in https://github.com/cda-tum/mqt-qecc/pull/390
- As part of the move to the new GitHub organization, we are updating the branding of the repositories and modernizing the documentation in that regard. A demo PR for this is here: https://github.com/munich-quantum-toolkit/core/pull/896 as well as the MQT Core main branch in general.
- Over in MQT Core, we now also adopted a CHANGELOG and and UPGRADE_GUIDE document to more clearly communicate changes to people. I believe this would be really good to have here as well: https://github.com/munich-quantum-toolkit/core/pull/909

That should cover it for now. I'm pretty sure I missed a couple of aspects, but tackling the above is already a great step forward.

burgholzer:
I also just pushed one more commit to this PR that raises the miminum required Qiskit version. This seems to be the first version that fixed the errors observed in the previous CI run.

PR #519, opened 2025-04-15
:fire: Dropped `TKET` dependency

simon1hofmann:


pyproject.toml
```diff
@@ -19,12 +19,9 @@ requires-python = ">=3.10"
 dynamic = ["version"]
 
 dependencies = [
-    #  manages the dependencies for qiskit and tket in a combined fashion
-    "pytket-qiskit>=0.53.0",
-    # lowest version that supports the used pytket AutoRebase pass instead of auto_rebase
-    "pytket>=1.29.0",
     # Qiskit 1.2.1 contains some fixes for importing OpenQASM 3 files
     "qiskit>=1.2.1",
+    "qiskit<=1.4.2",
```
burgholzer, line 24: You should be able to work around the build failures by replacing these lines with 
```
"qiskit>=1.2.1,<2"
```

pyproject.toml
```diff
@@ -19,12 +19,8 @@ requires-python = ">=3.10"
...
 
 dependencies = [
-    #  manages the dependencies for qiskit and tket in a combined fashion
-    "pytket-qiskit>=0.53.0",
-    # lowest version that supports the used pytket AutoRebase pass instead of auto_rebase
-    "pytket>=1.29.0",
     # Qiskit 1.2.1 contains some fixes for importing OpenQASM 3 files
-    "qiskit>=1.2.1",
+    "qiskit>=1.2.1,<2",
     "qiskit_ibm_runtime<0.37.0", # FakeBackends have been removed with 0.37.0
```
burgholzer, line 24: Ah. The error is actually coming from here, because there is no lower bound.
So the minimums session chooses the first published version.
You'd need to look up the first version on PyPI that supports Qiskit 1.2.1

Btw: you can also run these tests locally to debug
```
uvx nox -s minimums-3.12
```

simon1hofmann: I just checked the new minimal versions of packages that changed compared to `main`:                                                
qiskit-aer==0.1.0
qiskit-ibm-runtime==0.1.0
qiskit-terra==0.46.3 (new)
 
Previous minimal versions:
qiskit-aer==0.14.0.1
qiskit-ibm-runtime==0.23.0

tests/test_bench.py
```diff
@@ -370,28 +307,12 @@ def test_dj_constant_oracle() -> None:
...
-        compiler="tket",
-        provider_name="oqc",
-        device_name="oqc_lucy",
-    )
-    # check that all gates in the circuit are in the coupling map
-    cmap = [(c[0], c[1]) for c in OQCProvider.get_device("oqc_lucy").coupling_map]
-    assert qc.valid_connectivity(arch=pytket.architecture.Architecture(cmap), directed=True)
-
-
 @pytest.mark.parametrize(
```
nquetschlich, line 310: I could imagine that we can remove some of the tested benchmark instances. I used so many to cover all different combinations. Since tket is not available anymore, we most likely can reduce the number of test instances.

src/mqt/bench/cli.py
```diff
@@ -75,8 +70,6 @@ def main() -> None:
         print(qiskit_circuit_to_str(result))
         return
 
-    print(circuit_to_qasm_str(result))
```
nquetschlich, line 78: Do you know why this statement was here in the first place?

src/mqt/bench/cli.py
```diff
@@ -20,9 +19,7 @@ def format_help(self) -> str:
         """Include version information in the help message."""
         help_message = super().format_help()
         version_info = (
-            f"\nMQT Bench version: {metadata.version('mqt.bench')}\n"
-            f"Qiskit version: {metadata.version('qiskit')}\n"
-            f"TKET version: {metadata.version('pytket')}\n"
+            f"\nMQT Bench version: {metadata.version('mqt.bench')}\nQiskit version: {metadata.version('qiskit')}\n"
```
nquetschlich, line 22:
```suggestion
            f"\nMQT Bench version: {metadata.version('mqt.bench')}\n"
            f"Qiskit version: {metadata.version('qiskit')}\n"
```

src/mqt/bench/cli.py
simon1hofmann: I think it was a simplified if-else statement:

```
    if isinstance(result, QuantumCircuit):
        print(qiskit_circuit_to_str(result))
        return

(else)
    print(circuit_to_qasm_str(result))
```

src/mqt/bench/cli.py
simon1hofmann: `pre-commit` automatically formats it like this.

nquetschlich (approved):
Thanks for the changes. I think it makes sense that @burgholzer takes a look at this PR now.

burgholzer (approved):
This is looking great. Many thanks for all the work.
I merely pushed a couple of minor fix ups where I noticed that something was still slightly off.
Once CI is green. This is good to go in 🚀

PR #530, opened 2025-04-16
✨  Introduce `Hatchling` and `dependency-groups`

### LINKED ISSUE #524: ✨ Introduce PEP 735 dependency groups

MQT Bench currently does not use PEP 735 dependency groups. Inspiration for how to add these may come from MQT QECC, which recently switched to a pure-python package and has the most modern setup in that regard (see Remove C++ part of package. mqt-qecc#390)

see above

### LINKED ISSUE #525: ✨ Switch to `hatchling` as build backend

MQT Bench still uses setuptools as a build backend. There are better and more modern alternatives that also break less often. The concrete proposal here would be to switch to hatchling as a build backend. This is also covered in Remove C++ part of package. mqt-qecc#390

see above

nquetschlich:
resolved #524 and resolves #525


burgholzer (approved):
This LGTM. I just added a couple of finishing touches. Once CI is green, this should be good to go 🚀

PR #538, opened 2025-04-22
🔥 removed the generation logic 

nquetschlich:
Since in the (very near) future, benchmarks will be generated only through the CLI using the new frontend. Therefore, the old generation logic (that was mostly also not tested in the CI due to its time consumption) is removed.

### DIFF
--- a/src/mqt/bench/benchmark_generation.py
+++ b/src/mqt/bench/benchmark_generation.py
@@ -2,26 +2,21 @@
 
 from __future__ import annotations
 
-import argparse
-import json
 import signal
 import sys
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, TypedDict, overload
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


src/mqt/bench/benchmark_generation.py
```diff
@@ -2,26 +2,21 @@
...
 
-import argparse
-import json
 import signal
 import sys
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, TypedDict, overload
 from warnings import warn
 
-from joblib import Parallel, delayed
```
burgholzer, line 13: do we still need the `joblib` dependency if this is removed here?
(if not, please remove it and update the lock file with `uv lock`)

burgholzer (approved):
Removed code is debugged code. I like it 👍🏼 
One minor comment I was not 100% sure of.

burgholzer (approved):

PR #539, opened 2025-04-22
✨ refactored output logic

### LINKED ISSUE #523: ✨ Improve output format handling

We should generalize the output logic from purely OpenQASM (2 or 3) to more general output formats. Every function parameter that is called "qasm_format" should most likely be called "output_format" and should accept values from a string enum. 

Once a qiskit QuantumCircuit is the only immediate representation we are relying on, we can really streamline the output routines and move them to a dedicated module (no more utils 🙃). That module should, for the beginning, definitely contain routines for exporting to QASM2, QASM3, and Qiskit's QPY, which should be easy enough to support out of the box. 

In the future, I would imagine this to be extended by plugins, e.g., for tket, catalyst, cuda-q, mqt-core-ir, etc. The whole point is to decouple the generation logic from the output generation logic.

- Refactor the output logic
- Refactor the `utils.py` file by moving all output files to such a new module and move the remaining code to the generation module

nquetschlich:
This PR refactors the output logic. Resolves #523.

PR #538  should be merged first.


nquetschlich:
realized in #540

PR #540, opened 2025-04-22
♻️ refactored output module

nquetschlich:
resolves #539


src/mqt/bench/benchmark_generation.py
```diff
@@ -450,50 +444,97 @@ def get_benchmark(
...
+
+
+def get_default_qasm_output_path() -> str:
+    """Returns the path where all .qasm files are stored."""
+    return str(resources.files("mqt.bench") / "viewer" / "static" / "files" / "qasm_output")
+
+
+def get_zip_folder_path() -> str:
+    """Returns the path where the zip file is stored."""
+    return str(resources.files("mqt.bench") / "viewer" / "static" / "files")
```
burgholzer, line 486: More of a side comment unrelated to this PR, but that I noticed because these functions were moved over from the utils module as part of this PR:
These paths no longer seem right as the viewer no longer exists and the generated files do no longer belong inside the `mqt.bench` package.
Furthermore, the name `get_default_qasm_output_path` is no longer really suitable for the future, as we will not only be exporting QASM files (QPY being the easiest extension)

Maybe they aren't needed entirely after the generation logic has been dropped.

src/mqt/bench/output.py
burgholzer, line 1: Could we already make this slightly more modular?
In the future, this will not only be used for exporting QASM files.
The existing function already does two things that could be separated:
- getting a string representation of a quantum circuit program in some format
- generating a dedicated header for MQT Bench

We probably want the header in all different kinds of outputs in one way or another. So I think it makes sense to have this as a separate function.
In addition, I believe it should be possible to make this slightly more efficient by directly writing to a file (header first, then actual content) and avoiding serialising the program to a string.

Let's try to be as forward thinking as possible here to create a structure that we can easily extend in the future.
It probably makes sense to create some kind of StrEnum for the types of outputs we support.
I would guess that the header function needs to have that as input as well in order to know how to prepare the appropriate header for the respective program.
QPY would be good to incorporate here because it is a binary format that needs different handling than the text-based formats we currently support.

burgholzer (changes requested):
Thanks for pushing this forward!

I think this is already a good step in the right direction. However, let's take this a step further and make this ready for further future extensions.

src/mqt/bench/benchmark_generation.py
burgholzer: Great 👍🏼 then let's remove them

src/mqt/bench/cli.py
```diff
@@ -23,56 +26,127 @@ def format_help(self) -> str:
...
+        output_format=str(fmt.value),
+        target_directory=args.target_directory,
+    )
+    if not success:
+        sys.exit(1)
+    # Optionally, inform user of file location if saving
+    if args.save or fmt == OutputFormat.QPY:
+        ext = "qasm" if fmt in (OutputFormat.QASM2, OutputFormat.QASM3) else fmt.value
+        path = Path(args.target_directory) / f"{filename}.{ext}"
+        print(path)
```
burgholzer, line 148: It would be nice to not replicate too much logic here.
Can we somehow centralise this in the `output.py` file?
This is not so different to writing to a file. It just writes to stdout instead.

src/mqt/bench/cli.py
```diff
@@ -23,56 +26,127 @@ def format_help(self) -> str:
...
+    parser.add_argument(
+        "--device",
+        type=str,
+        help="Device name for mapping stage (e.g., 'ibm_washington').",
+    )
+    parser.add_argument(
+        "--output-format",
+        type=str,
+        choices=[fmt.value for fmt in OutputFormat],
+        default=OutputFormat.QASM2.value,
```
burgholzer, line 78:
```suggestion
        default=OutputFormat.QASM3.value,
```
probably more future proof

src/mqt/bench/cli.py
```diff
@@ -23,56 +26,127 @@ def format_help(self) -> str:
...
+        "--target-filename",
+        type=str,
+        default="",
+        help="Base filename (without extension) for saved output.",
+    )
+    parser.add_argument(
+        "--save",
+        action="store_true",
+        help="If set, save the output to a file instead of printing to stdout (required for 'qpy').",
+    )
```
burgholzer, line 97: I like this addition of writing to a file from the CLI instead of printing to stdout.
I'd probably slightly rephrase the existing help text for the `save` flag to just mention `qpy` as one of the examples that is not available as text-based output.
Additionally, I do not think that the `--target-filename` argument is necessary or helpful, but maybe I am overlooking things. MQT Bench has its naming scheme for files and I am not sure the user should be allowed to mess with it.
Of course, there needs to be some way for the user to specify where things should get stored.

tests/test_cli.py
```diff
@@ -116,3 +102,40 @@ def test_cli_errors(args: list[str], expected_output: str, script_runner: Script
...
+        "--target-directory", target_dir,
+        "--target-filename", target_file,
+    ])
+    # File should exist
+    assert ret_qpy.success
+    # Should output the path to the saved file
+    expected_path = str(Path(target_dir) / f"{target_file}.qpy")
+    assert expected_path in ret_qpy.stdout
+    # File should exist
+    assert (tmp_path / f"{target_file}.qpy").is_file()
```
burgholzer, line 141: Might be better as individual smaller tests to kind of keep the spirit of a "unit" test.

burgholzer (changes requested):
Many thanks for all the work on this!
I really like where this is going.
You'll find detailed feedback in the inline comments.
Most likely, this will be good to go after this feedback round.

simon1hofmann:
> Many thanks for all the work on this! I really like where this is going. You'll find detailed feedback in the inline comments. Most likely, this will be good to go after this feedback round.

Thanks a lot for the detailed and helpful review, I tried to include all comments and suggestions.

burgholzer (approved):
Nice work, this is looking great!
I especially like the thoroughness of the added tests!

PR #549, opened 2025-04-30
:art: Removed `Provider` and improved `Device` Structure

### LINKED ISSUE #387: Remove the `Provider` concept and replace it with base classes of `Device`s

simon1hofmann:
Extracted from #429:
1 removes the Provider class (resolves #387)
4 adds a new IBM device: ibmq_torino (and, with that, resolves https://github.com/munich-quantum-toolkit/bench/issues/220 kind of)
5 improves the Device class structure
6 refactors the tests to accelerate the test execution time


nquetschlich (approved):
Thanks a lot for refactoring this PR, @simon1hofmann. 

LGTM so far. Before merging, we should check in detail again, whether the number of test cases in the `test_bench.py` file is as small as possible to save CI runtime. 

I think it makes sense that @burgholzer takes a look now.

src/mqt/bench/devices/ibm.py
```diff
@@ -10,136 +10,58 @@
...
-        device.basis_gates = ibm_calibration["basis_gates"]
-        device.coupling_map = list(ibm_calibration["connectivity"])
+    def __init__(self, backend: BackendV2, device_name: str, gateset_name: str) -> None:
+        """Initialize the IBM device."""
+        self.name = device_name
+        self.gateset = Gateset(gateset_name, backend.operation_names)
+        self.num_qubits = backend.num_qubits
+        self.coupling_map = backend.coupling_map
+        self.calibration = None
+        self.target = backend.target
```
burgholzer, line 38: That's probably the best sign that the device class itself is (or at least many of its members are) redundant. We already store the target here and that contains practically all information.

src/mqt/bench/devices/quantinuum.py
```diff
@@ -11,14 +11,41 @@
...
+    def __init__(self) -> None:
+        """Initialize the Quantinuum device."""
+        with get_device_calibration_path("quantinuum_h2").open() as json_file:
+            self.quantinuum_calibration = cast("QuantinuumCalibration", json.load(json_file))
+        self.calibration = None
+
+        self.name = self.quantinuum_calibration["name"]
+        self.gateset = Gateset("quantinuum", self.quantinuum_calibration["basis_gates"])
+        self.num_qubits = self.quantinuum_calibration["num_qubits"]
+        self.coupling_map = list(self.quantinuum_calibration["connectivity"])
```
burgholzer, line 32: When we switch to Qiskit's Target for representing devices, we can still use JSON files for representing the devices. We just need to follow IBMs format.

This is probably a bit out of place here, but we probably also want to vendor the Qiskit Runtime-provided fake backend data as part of MQT Bench so that we can remove the ibm-Qiskit-runtime dependency from MQT Bench for now.

src/mqt/bench/cli.py
```diff
@@ -68,9 +68,9 @@ def main() -> None:
         help="Qiskit compiler optimization level (0-3).",
     )
     parser.add_argument(
-        "--native-gate-set",
+        "--gateset",
         type=str,
-        help="Provider name for native gate set (e.g., 'ibm', 'rigetti').",
+        help="Used gate set (e.g., 'ibm', 'rigetti').",
```
burgholzer, line 73: I might be wrong, but I don't think there is an `ibm` gateset anymore, right?

src/mqt/bench/benchmark_generation.py
```diff
@@ -72,7 +67,7 @@ def generate_filename(
     benchmark_name: str,
     level: str,
     num_qubits: int | None,
-    provider_name: str | None = None,
+    gateset_name: str | None = None,
     device_name: str | None = None,
```
burgholzer, line 71: I think this interface could be typed a little stronger.
Instead of strings for the gateset and the device name, I'd rather pass a `Gateset` and a `Device` object here.

burgholzer (changes requested):
Many thanks for pulling this PR together 🙏 great to see this moving forward!

You'll find some detailed comments inline.
After getting through all the individual files, I kept thinking about the Provider/Gateset/Device handling and the potential move to Qiskit's `Target`.
That led me to the following idea: what if we do not only replace the existing device class with the Target, but also remove the dedicated "gateset" parameter from essentially all methods and functions.
A "Target" already fully describes the device and its gateset. So if one wants a circuit for a particular gateset, one can also simply specify a respective target/device.
This would considerably simplify the overall interface. And, for convenience, we could introduce generic providers for a specific gateset that take an optional parameter in their constructor for the number of qubits. That way, the compilation at the native gates level is not limited to the number of qubits offered by a real device.

The benchmark generation function would simply take a `Target`. The individual devices can be implemented as derived classes from the `Target` class.
Does that make sense? If so, it would be great to see it in action.

Note that, at the mapped level, the `transpile` call can be adjusted to simply take the `target` as an argument.

src/mqt/bench/devices/calibration.py
```diff
@@ -243,3 +244,14 @@ def compute_average_readout_duration(self) -> float:
...
+
+def get_device_calibration_path(filename: str) -> Path:
+    """Get the path to the calibration file for a device."""
+    calibration_path = resources.files("mqt.bench") / "calibration_files" / f"{filename}_calibration.json"
+
+    if not calibration_path.is_file():
+        msg = f"Calibration file not found: {calibration_path}"
+        raise FileNotFoundError(msg)
+
+    return Path(str(calibration_path))
```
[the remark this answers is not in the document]
simon1hofmann: Only the calibration files for `ibm` devices were removed, for the others (e.g. `ionq`) this function is still used.

src/mqt/bench/devices/device.py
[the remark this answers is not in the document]
simon1hofmann: As the other comments are quite easy to address, maybe we can merge this PR and do the replacement with Qiskit's `Target` class in a separate PR? 
Just want to avoid this PR getting too big again.

simon1hofmann:
> LGTM 👍🏼 And I fully agree that tackling the potential move to Qiskit's Target is better handled in a separate PR. I'll leave the respective conversations open here. @simon1hofmann can you please take note of them, resolve them, and then merge this PR?

Thanks a lot, I moved your comments to this feature request: #553

src/mqt/bench/devices/__init__.py
```diff
@@ -14,66 +14,35 @@
...
+        IBMTorino(),
+        IBMMontreal(),
+        IBMWashington(),
+        IonQAria1(),
+        IonQHarmony(),
+        IQMAdonis(),
+        IQMApollo(),
+        OQCLucy(),
+        QuantinuumH2(),
+        RigettiAspenM3(),
```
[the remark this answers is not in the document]
simon1hofmann: Moved to #554

Issue #553, opened 2025-05-05
✨ Replace devices with Qiskit `Target``s


simon1hofmann:
### What's the problem this feature will solve?

Ideas from #549:
That led me to the following idea: what if we do not only replace the existing device class with the `Target`, but also remove the dedicated `gateset` parameter from essentially all methods and functions.
A `Target` already fully describes the device and its gateset. So if one wants a circuit for a particular gateset, one can also simply specify a respective target/device.
This would considerably simplify the overall interface. And, for convenience, we could introduce generic providers for a specific gateset that take an optional parameter in their constructor for the number of qubits. That way, the compilation at the native gates level is not limited to the number of qubits offered by a real device.

The benchmark generation function would simply take a `Target`. The individual devices can be implemented as derived classes from the `Target` class.

Note that, at the mapped level, the `transpile` call can be adjusted to simply take the `Target` as an argument.

When we switch to Qiskit's `Target` for representing devices, we can still use JSON files for representing the devices. We just need to follow IBMs format.

### Describe the solution you'd like

Switch to `Target`'s.

PR #560, opened 2025-05-13
✨  Introducing Qiskit's Target as our main Device representation and Remove OQC Lucy

nquetschlich:
This PR introduces the Qiskit Target object as our new device representation (resolves issue #553).

Furthermore, the deprecated OQC Lucy device was removed.


src/mqt/bench/benchmark_generation.py
```diff
@@ -69,8 +69,8 @@ def generate_filename(
     benchmark_name: str,
     level: str,
     num_qubits: int | None,
-    gateset: Gateset | None = None,
-    device: Device | None = None,
+    gateset: Target | None = None,
+    device: Target | None = None,
```
burgholzer, line 73: There is no need to distinguish between these anymore.
It is totally sufficient to have a single `target: Target | None = None` parameter.
The native gates level will just extract the gateset from the target.
The mapped level will use the full target.
The other levels do not use the information at all.

src/mqt/bench/benchmark_generation.py
```diff
@@ -90,10 +90,10 @@ def generate_filename(
     base = f"{benchmark_name}_{level}"
 
     if level == "nativegates" and gateset is not None and opt_level is not None:
-        return f"{base}_{gateset.name}_opt{opt_level}_{num_qubits}"
+        return f"{base}_{gateset.description}_opt{opt_level}_{num_qubits}"
```
burgholzer, line 93: I am fairly sure the description needs some kind of sanitization.
Many descriptions of targets in the wild will use spaces (`"Hardware Provider XZY Target"`). 
Based on the above comment, the gateset should be a serialized/stringified version of the gateset and the device description should be a sanitized version of the description.

src/mqt/bench/benchmark_generation.py
```diff
@@ -219,7 +219,7 @@ def get_indep_level(
 @overload
 def get_native_gates_level(
     qc: QuantumCircuit,
-    gateset: Gateset,
+    gateset: Target,
```
burgholzer, line 222: I'd simply refer to this as `target` everywhere.

src/mqt/bench/benchmark_generation.py
```diff
@@ -284,14 +284,16 @@ def get_native_gates_level(
...
         return True
 
-    if gateset.name == "clifford+t":
+    if gateset.description == "clifford+t":
         from qiskit.transpiler import PassManager  # noqa: PLC0415
         from qiskit.transpiler.passes.synthesis import SolovayKitaev  # noqa: PLC0415
 
         # Transpile the circuit to single- and two-qubit gates including rotations
+        gates = {inst.name for inst, _ in gateset.instructions}
+        print(gates)
```
burgholzer, line 293: There is a print statement here, which I am pretty surprised that ruff did not complain about or optimize away. Maybe worth checking the config.

Other than that, we should simply be able to use `*target.operation_names` in the call down below.

src/mqt/bench/benchmark_generation.py
```diff
@@ -303,12 +305,12 @@ def get_native_gates_level(
...
         compiled = transpile(
             new_qc,
-            basis_gates=gateset.gates,
+            target=gateset,
             optimization_level=opt_level,
             seed_transpiler=10,
         )
     else:
-        compiled = transpile(qc, basis_gates=gateset.gates, optimization_level=opt_level, seed_transpiler=10)
+        compiled = transpile(qc, target=gateset, optimization_level=opt_level, seed_transpiler=10)
```
burgholzer, line 313: I am fairly sure that we only want to pass the `target.operation_names` to the `basis_gates` kwarg here for the native gates level.
If a regular target were to be passed in, we would also consider its connectivity here.
Please also check this for every other place where the native gates level is used.

src/mqt/bench/benchmark_generation.py
```diff
@@ -414,7 +413,7 @@ def get_benchmark(
     circuit_size: int | None = None,
     benchmark_instance_name: str | None = None,
     compiler_settings: CompilerSettings | None = None,
-    gateset: str | Gateset = "ibm_falcon",
+    gateset: str | Target = "ibm_falcon",
     device_name: str = "ibm_washington",
```
burgholzer, line 417: Just to be sure. This should also become one parameter here.
And I am fairly sure, that we do not want to allow strings here anymore, but only real `Target`s.
We can provide a separate convenience method to instantiate targets by name, but I would argue it helps the interface to reduce the potential typing differences.

src/mqt/bench/calibration_files/oqc_lucy_calibration.json
burgholzer, line 1: OQC does have a new device that would be interesting to add as a substitute. (https://oqc.tech/tech/toshiko/)
But let's maybe reserve this for a follow up PR.

src/mqt/bench/devices/__init__.py
```diff
@@ -6,131 +6,206 @@
...
+            elif gate.num_qubits == 2:
+                qargs = {(q0, q1): None for q0 in range(num_qubits) for q1 in range(num_qubits) if q0 != q1}
+            else:
+                continue  # skip 3+ qubit gates (not needed here)
+
+            target.add_instruction(gate, qargs)
+        except Exception:
+            continue
+
+    return target
```
burgholzer, line 140: This should definitely be moved to a separate file.
Most importantly, this is not just "clifford" but "clifford_t"

src/mqt/bench/devices/__init__.py
```diff
@@ -6,131 +6,206 @@
...
+            elif gate.num_qubits == 1:
+                qargs = {(q,): None for q in range(num_qubits)}
+            elif gate.num_qubits == 2:
+                qargs = {(q0, q1): None for q0 in range(num_qubits) for q1 in range(num_qubits) if q0 != q1}
+            else:
+                continue  # skip 3+ qubit gates (not needed here)
+
+            target.add_instruction(gate, qargs)
+        except Exception:
+            continue
```
burgholzer, line 138: This can be much simpler I believe.
As per the docstring of the `Target.add_instruction` method, there is no need to supply properties if there aren't any.
So you also do not need the dictionary on top.
Simply call `target.add_instruction(Gate())` for each of the items in the list and that is it.

src/mqt/bench/devices/__init__.py
```diff
@@ -6,131 +6,206 @@
...
+    return target
 
 
 @cache
-def get_available_native_gatesets() -> list[Gateset]:
-    """Get a list of all available native gatesets."""
-    available_gatesets = []
+def get_available_native_gatesets(num_qubits: int = 20) -> dict[str, Target]:
+    """Return mapping: gateset_name → Target (deduplicated)."""
+    gatesets: dict[str, Target] = {}
```
burgholzer, line 146: Instead of doing it this way (the whole function), I think I would much rather just define generic backends for the individual gate-sets that we support. The individual providers can all expose one generic target for each of the different gate-sets they expose. This should be extremely straight-forward and much more fool-proof then the logic here.

src/mqt/bench/devices/__init__.py
```diff
@@ -6,131 +6,206 @@
...
+            continue
+
+    return target
 
 
 @cache
-def get_available_native_gatesets() -> list[Gateset]:
-    """Get a list of all available native gatesets."""
-    available_gatesets = []
+def get_available_native_gatesets(num_qubits: int = 20) -> dict[str, Target]:
```
burgholzer, line 144: Based on the other comment of only using targets anymore, I am not quite sure if we even need this function here. There probably needs to be some kind of convenience function somewhere for getting the gatesets, but let's see.

src/mqt/bench/devices/calibration.py
burgholzer, line 1: I would argue that this file can be removed as I believe there should be no need to carry all these calibration files with us here.

src/mqt/bench/devices/ionq.py
```diff
@@ -6,91 +6,75 @@
...
 
-"""IonQ devices."""
+"""File to create a target device from the IonQ calibration data."""
+# Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+# Copyright (c) 2025 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
```
burgholzer, line 16: What happened here?
```suggestion
```
Please also fix in the other files.

src/mqt/bench/devices/ionq.py
```diff
@@ -6,91 +6,75 @@
...
-        """Initialize the IonQ Aria1 device."""
-        super().__init__(get_device_calibration_path("ionq_aria1"))
+    # Gate durations and fidelities
+    oneq_fidelity = calib["fidelity"]["1q"]["mean"]
+    twoq_fidelity = calib["fidelity"]["2q"]["mean"]
+    spam_fidelity = calib["fidelity"]["spam"]["mean"]
 
+    oneq_duration = calib["timing"]["1q"]
+    twoq_duration = calib["timing"]["2q"]
+    readout_duration = calib["timing"]["readout"]
```
burgholzer, line 49: I would be very tempted to simply hard code the calibration file values here.
We have never updated them. Since their creation. And they are fairly specific to a particular device.
We could simply link to the origin of the data.
Or, wild and crazy idea: Simply remove all the calibration information for now to streamline the implementation and then come up with a good strategy to selectively add it back wherever feasible.
MQT Bench was not using that information so far anyway.

src/mqt/bench/output.py
```diff
@@ -86,8 +85,10 @@ def generate_header(
         f"// Output format: {fmt.value}",
     ))
 
-    if gateset:
+    if target:
+        gateset = sorted({str(inst.name) for inst, _ in target.instructions})
```
burgholzer, line 89:
```suggestion
    if target:
        gateset = target.operation_names
```
or similar

src/mqt/bench/output.py
```diff
@@ -86,8 +85,10 @@ def generate_header(
...
         lines.append(f"// Used gateset: {gateset}")
+    c_map = target.build_coupling_map() if target else None
```
burgholzer, line 91: Pretty sure this function will need an additional `level` input.
Because otherwise, this will add a coupling map even if a circuit is only transpiled to the native gates level

burgholzer (changes requested):
Thanks @nquetschlich for spearheading this 🙏🏼 
I have not made it through the whole PR yet, but I think I got enough of an idea to suggest targeted (pun intended) improvements.
I hope the inline comments roughly make sense and give you some direction on where this could/should be heading.

src/mqt/bench/benchmark_generation.py
nquetschlich: Agreed.

src/mqt/bench/benchmark_generation.py
nquetschlich: Agreed and adjusted it. However, for the `get_benchmark` method I think it is more easily understandable to still call it `gateset`. I would like to discuss this and some other points in person.

src/mqt/bench/benchmark_generation.py
nquetschlich: Thanks for catching that, your suggestions works well.

src/mqt/bench/devices/__init__.py
nquetschlich: Good catch, I created separate files for both the native gateset and the device handling.

src/mqt/bench/devices/ionq.py
nquetschlich: I like the idea and would like to discuss the consequences in terms of MQT Predictor in person. Let's then decide based on that discussion.

src/mqt/bench/output.py
nquetschlich: Thanks, followed that suggestion.

src/mqt/bench/output.py
nquetschlich: I don't think so, since the `build_coupling_map()` functions returns `None` according to its docstring if there are no connectivity constraints present in the target (see https://github.com/Qiskit/qiskit/blob/ee4095933e688efb794e679b3feda658fefdec9b/qiskit/transpiler/target.py#L545).

src/mqt/bench/devices/__init__.py
nquetschlich: I re-structured the code to make it more modular.

burgholzer:
Many thanks for your work on this. 🙏🏼 
I pushed a couple of commits that simply, refactor, and add a couple of things.
I'd love to work more on this, but I have to stop now and get some sleep.

I think this is a pretty solid improvement already.
What I noticed though is that some of the devices are either outdated (Rigetti for example) or their native gate sets are simply not correct (IonQ; compare with https://github.com/qiskit-community/qiskit-ionq/blob/main/qiskit_ionq/ionq_gates.py).
I tried to add a couple more devices to the IBM and IQM side of things, but do not have the time to touch the other providers.
It would be good to work on the other device representations as a follow-up to this PR so that they more closely match the current state of the field.

Let's see if CI passes and what you think about some of the changes here.

PR #567, opened 2025-05-21
:memo: Added CHANGELOG and UPGRADING information

simon1hofmann:
This PR adds a dedicated changelog and upgrade guide to the project as a centralized place to keep track of significant updates to this library.
Both of these files are curated selections of the PRs that go into a release.
Notably, these are written for humans and not for machines.
As such, they are not automatically scraped together like the drafts created by the release drafter action. These drafts may still serve as blueprints for corresponding changelog entries.


simon1hofmann:
@nquetschlich I created a first draft of the changelog based on the release draft, but we might need to rewrite that based on what is really necessary. 

Also, in the release draft only my name is mentioned even though you co-authored it. Should we include all authors in the `CHANGELOG`?

simon1hofmann:
@burgholzer This PR now only contains a skeleton for the `CHANGELOG´ and `UPGRADING` to be completed when releasing `v2.0.0`.

burgholzer (approved):
LGTM 👍🏼
Thanks for setting this up!

PR #569, opened 2025-05-22
🚨Fixed Pending Deprecation Warnings

simon1hofmann:
Fixed warnings related to pending deprecation as of Qiskit `v1.3.x`


src/mqt/bench/benchmarks/ae.py
```diff
@@ -11,8 +11,18 @@
...
 import numpy as np
+from packaging.version import Version
+from qiskit import version
 from qiskit.circuit import QuantumCircuit
-from qiskit.circuit.library import PhaseEstimation
+
+if Version(version.get_version_info()) >= Version("1.3.2"):
+    from qiskit.circuit.library import phase_estimation
+else:
+    from qiskit.circuit.library import PhaseEstimation
```
burgholzer, line 21: I'd slightly prefer the Pythonic "try and except" pattern that does not require us to import packaging as well as the `version` module from qiskit.
This, and all the other similar statements, can simply become
```suggestion
try
    from qiskit.circuit.library import phase_estimation
except ImportError:
    from qiskit.circuit.library import PhaseEstimation
```

docs/benchmark_selection.md
burgholzer, line 1: This is something to keep in mind for the changelog. We are effectively dropping two circuit variants here. This should be explicitly highlighted.

src/mqt/bench/benchmarks/shor.py
```diff
@@ -31,7 +31,8 @@
 
 import numpy as np
 from qiskit.circuit import Instruction, ParameterVector, QuantumCircuit, QuantumRegister
-from qiskit.circuit.library import QFT
+from qiskit.circuit.library import QFTGate
+from qiskit.synthesis import synth_qft_full
```
burgholzer, line 35: Isn't it slightly inconsistent to use both of these methods here instead of just one of them? Also holds for other places where either one of these is used.

src/mqt/bench/benchmarks/ae.py
simon1hofmann: This leads to `TypeError: 'module' object is not callable` when using Qiskit `v1.2.1`, probably because the file is called `phase_esimation.py`?

src/mqt/bench/benchmarks/shor.py
burgholzer: Ah. In that case I think it might actually make sense to leave it as is and use `QFTGate` wherever that is enough and only use `synth_qft_full` when needed.
There are some high-level synthesis plugins that might take the QFT Gate and synthesize it differently. I remember reading something about that.

src/mqt/bench/benchmarks/ae.py
simon1hofmann: For the except block, should we go with:

```
from qiskit.circuit.library import PhaseEstimation

    def phase_estimation(num_eval: int, grover: QuantumCircuit) -> PhaseEstimation:
        """PhaseEstimation (Qiskit < 1.3.2)."""
        return PhaseEstimation(num_eval, grover)
```
        
or just ignore the `N813 Camelcase` warning?

`from qiskit.circuit.library import PhaseEstimation as phase_estimation  # noqa: N813`

PR #570, opened 2025-05-22
🎨 IonQ Device Update

nquetschlich:
This PR...
- updates the IonQ device calibration data
- introduces a new device: IonQ Forte 1
- introduces a more correct gateset for these devices

### DIFF
--- a/src/mqt/bench/targets/devices/__init__.py
+++ b/src/mqt/bench/targets/devices/__init__.py
@@ -14,7 +14,7 @@ from functools import cache
 from typing import TYPE_CHECKING
 
 from .ibm import get_ibm_target
-from .ionq import get_ionq_target
+from .ionq import add_equivalences, get_ionq_target
 from .iqm import get_iqm_target
 from .quantinuum import get_quantinuum_target
 from .rigetti import get_rigetti_target
@@ -24,6 +24,7 @@ if TYPE_CHECKING:
 
 
 __all__ = [
+    "add_equivalences",
     "get_available_devices",
     "get_device_by_name",
 ]
@@ -38,8 +39,8 @@ def get_available_devices() -> list[Target]:
         get_ibm_target("ibm_eagle_127"),
         get_ibm_target("ibm_heron_133"),
         get_ibm_target("ibm_heron_156"),
-        get_ionq_target("ionq_harmony"),
-        get_ionq_target("ionq_aria1"),
+        get_ionq_target("ionq_forte_1"),
+        get_ionq_target("ionq_aria_1"),
         get_iqm_target("iqm_crystal_5"),
         get_iqm_target("iqm_crystal_20"),
         get_iqm_target("iqm_crystal_54"),
--- a/src/mqt/bench/targets/devices/ionq.py
+++ b/src/mqt/bench/targets/devices/ionq.py
@@ -10,48 +10,382 @@
 
 from __future__ import annotations
 
-from qiskit.circuit import Parameter
-from qiskit.circuit.library import Measure, RXGate, RXXGate, RYGate, RZGate
+import math
+from typing import TYPE_CHECKING
+
+import numpy as np
+from qiskit import QuantumRegister
+from qiskit.circuit import Gate, Parameter, QuantumCircuit
+from qiskit.circuit.equivalence_library import SessionEquivalenceLibrary
+from qiskit.circuit.library import CU3Gate, CXGate, Measure, RXGate, RZGate, UGate, XGate
 from qiskit.transpiler import InstructionProperties, Target
+
+if TYPE_CHECKING:
+    from qiskit.circuit.parameterexpression import ParameterValueType
+
+
+class GPIGate(Gate):
+    r"""Single-qubit GPI gate.
+    **Circuit symbol:**
+    .. parsed-literal::
+             ┌───────┐
+        q_0: ┤ GPI(φ)├
+             └───────┘
+    **Matrix Representation:**.
+
+    .. math::
+
+       GPI(\phi) =
+            \begin{pmatrix}
+                0 & e^{-i*2*\pi*\phi} \\
+                e^{i*2*\pi*\phi} & 0
+            \end{pmatrix}
+    """
+
+    def __init__(self, phi: ParameterValueType, label: str | None = None) -> None:
+        """Create new GPI gate."""
+        super().__init__("gpi", 1, [phi], label=label)
+        q = QuantumRegister(1, "q")
+        qc = QuantumCircuit(q)
+        qc.x(0)
+        qc.rz(4 * phi * np.pi, 0)
+
+        self.definition = qc
+
+    def __array__(self, dtype: np.dtype[np.complex128] | None = None) -> np.ndarray:
+        """Return a numpy array for the GPI gate."""
+        top = np.exp(-1j * 2 * math.pi * self.params[0])
+        bottom = np.exp(1j * 2 * math.pi * self.params[0])
+        return np.array([[0, top], [bottom, 0]], dtype=dtype)
+
+
+class GPI2Gate(Gate):
+    r"""Single-qubit GPI2 gate.
+    **Circuit symbol:**
+    .. parsed-literal::
+             ┌───────┐
+        q_0: ┤GPI2(φ)├
+             └───────┘
+    **Matrix Representation:**.
+
+    .. math::
+
+        GPI2(\phi) =
+            \frac{1}{\sqrt{2}}
+            \begin{pmatrix}
+                1 & -i*e^{-i*2*\pi*\phi} \\
+                -i*e^{i*2*\pi*\phi} & 1
+            \end{pmatrix}
+    """
+
+    def __init__(self, phi: ParameterValueType, label: str | None = None) -> None:
+        """Create new GPI2 gate."""
+        super().__init__("gpi2", 1, [phi], label=label)
+        q = QuantumRegister(1, "q")
+        qc = QuantumCircuit(q)
+        qc.rz(-2 * phi * np.pi, 0)
+        qc.rx(np.pi / 2, 0)
+        qc.rz(2 * phi * np.pi, 0)
+
+        self.definition = qc
+
+    def __array__(self, dtype: np.dtype[np.complex128] | None = None) -> np.ndarray:
+        """Return a numpy array for the GPI2 gate."""
+        top = -1j * np.exp(-1j * self.params[0] * 2 * math.pi)
+        bottom = -1j * np.exp(1j * self.params[0] * 2 * math.pi)
+        return 1 / np.sqrt(2) * np.array([[1, top], [bottom, 1]], dtype=dtype)
+
+
+class MSGate(Gate):
+    r"""Entangling 2-Qubit MS gate.
+    **Circuit symbol:**
+    .. parsed-literal::
+              _______
+        q_0: ┤       ├-
+             |MS(ϴ,0)|
+        q_1: ┤       ├-
+             └───────┘
+    **Matrix representation:**.
+
+    .. math::
+
+       MS(\phi_0, \phi_1, \theta) =
+            \begin{pmatrix}
+                cos(\theta*\pi) & 0 & 0 & -i*e^{-i*2*\pi(\phi_0+\phi_1)}*sin(\theta*\pi) \\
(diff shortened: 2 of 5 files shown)


src/mqt/bench/targets/devices/ionq.py
```diff
@@ -10,55 +10,394 @@
...
+
+def zz_gate_equivalence() -> None:
+    """Add ZZ gate equivalence to the SessionEquivalenceLibrary."""
+    q = QuantumRegister(2, "q")
+    theta_param = Parameter("theta_param")
+    zz_circuit = QuantumCircuit(q)
+    zz_circuit.cx(0, 1)
+    zz_circuit.rz(2 * np.pi * theta_param, 1)
+    zz_circuit.cx(0, 1)
+    SessionEquivalenceLibrary.add_equivalence(ZZGate(theta_param), zz_circuit)
```
burgholzer, line 337: I am pretty sure these are not needed for our purposes and might actually be covered by the `_define` or the the `__array__` method.

src/mqt/bench/targets/devices/ionq.py
```diff
@@ -10,55 +10,394 @@
...
+
+    def __init__(self, phi: ParameterValueType, label: str | None = None) -> None:
+        """Create new GPI gate."""
+        super().__init__("gpi", 1, [phi], label=label)
+        q = QuantumRegister(1, "q")
+        qc = QuantumCircuit(q)
+        qc.x(0)
+        qc.rz(4 * phi * np.pi, 0)
+
+        self.definition = qc
```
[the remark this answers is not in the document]
nquetschlich: I implemented it now accordingly.

src/mqt/bench/targets/devices/ionq.py
```diff
@@ -10,55 +10,394 @@
...
+                0 & -i*e^{-i*2*\pi(\phi_0-\phi_1)}*sin(\theta*\pi) & cos(\theta*\pi) & 0 \\
+                -i*e^{i*2*\pi(\phi_0+\phi_1)}*sin(\theta*\pi) & 0 & 0 & cos(\theta*\pi)
+            \end{pmatrix}
+    """
+
+    def __init__(
+        self,
+        phi0: ParameterValueType,
+        phi1: ParameterValueType,
+        theta: ParameterValueType,
```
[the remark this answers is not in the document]
nquetschlich: Implemented it.

src/mqt/bench/targets/devices/ionq.py
nquetschlich: I removed these methods.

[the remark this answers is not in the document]
nquetschlich: Implemented it.

src/mqt/bench/targets/devices/__init__.py
```diff
@@ -38,8 +39,8 @@ def get_available_devices() -> list[Target]:
         get_ibm_target("ibm_eagle_127"),
         get_ibm_target("ibm_heron_133"),
         get_ibm_target("ibm_heron_156"),
-        get_ionq_target("ionq_harmony"),
-        get_ionq_target("ionq_aria1"),
+        get_ionq_target("ionq_forte_1"),
+        get_ionq_target("ionq_aria_1"),
```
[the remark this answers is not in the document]
nquetschlich: These are the specific device names and there is also an `Aria 2`. Therefore, I suggest to keep the name, but this is independent of the qubit number at the end - if you prefer that, I am happy to add it to all devices.

src/mqt/bench/targets/devices/ionq.py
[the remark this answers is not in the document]
nquetschlich: I by now restructured the module and implemented must of the mentioned things. However, two are a bit unclear to me:
- Regarding the re-write rules: Do you have a concrete idea how to implement the custom optimization passes? Should we in the compile routine check what target is used and then associate a specific passmanager with each target?
- Currently, only IBM uses the respectively defined gateset explicitly, since it does *not* attach calibration data to the respective operations. Do you have an idea how to do it similarly for all the other devices *with* calibration data?

src/mqt/bench/targets/devices/__init__.py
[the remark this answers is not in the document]
burgholzer: But aria 2 still uses the same layout and the same gateset if I am not mistaken.
Given how the device data is averaged anyway, I do not think there is too much value to this number.

If the gateset should really be different between these, then I would actually add the number to the gateset name and still append an underscore with the number of qubits for the device here.

Same holds for the Rigetti device.

src/mqt/bench/targets/devices/ionq.py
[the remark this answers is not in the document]
burgholzer: Regarding rewrite rules: these would indeed need to be added to device (or at least gateset) specific pass managers.
Definitely nothing for this PR but a (reasonable) follow up.

Regarding gateset: good question. Not too easy to realize that at the moment. But the current code looks fine as it is to me.

src/mqt/bench/targets/gatesets/ionq.py
```diff
@@ -0,0 +1,283 @@
...
+    lambda_param = Parameter("lambda_param")
+    u_gate = QuantumCircuit(q)
+    # this sequence can be compacted if virtual-z gates will be introduced
+    u_gate.append(GPI2Gate(0.5 - lambda_param / (2 * np.pi)), [0])
+    u_gate.append(
+        GPIGate(theta_param / (4 * np.pi) + phi_param / (4 * np.pi) - lambda_param / (4 * np.pi)),
+        [0],
+    )
+    u_gate.append(GPI2Gate(0.5 + phi_param / (2 * np.pi)), [0])
+    SessionEquivalenceLibrary.add_equivalence(UGate(theta_param, phi_param, lambda_param), u_gate)
```
burgholzer, line 253: I believe it would be better if these methods would take an equivalence library as a parameter instead of unconditionally adding to the SEL.
Gives a bit more flexibility.

src/mqt/bench/targets/gatesets/rigetti.py
```diff
@@ -0,0 +1,173 @@
...
+    qc.append(RZGate(phi), [qr[0]])
+    qc.append(RXMinusPiOver2Gate(), [qr[0]])
+    qc.append(RZGate(theta), [qr[0]])
+    qc.append(RXPiOver2Gate(), [qr[0]])
+    qc.append(RZGate(lam), [qr[0]])
+
+    SessionEquivalenceLibrary.add_equivalence(UGate(theta, phi, lam), qc)
+
+
+def cx_gate_equivalence() -> None:
```
burgholzer, line 141: Is it really true that a single CX gate takes two iSWAP gates? Seems counterintuitive to me.
Furthermore, the iSWAP gate is a Qiskit standard gate, so this equivalence should not really be needed.

src/mqt/bench/targets/gatesets/rigetti.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _define(self) -> None:
+        """Define RX(-π/2) gate using standard RX."""
+        q = QuantumRegister(1, "q")
+        qc = QuantumCircuit(q)
+        qc.rx(-np.pi / 2, 0)
+        self.definition = qc
+
+
+def u_gate_equivalence() -> None:
```
burgholzer, line 122: I think it could be worth to add a couple more equivalences here that take fewer gates.
Equivalences for the real (parametrized) RX gate come to mind.
Pretty sure there are others.

burgholzer (changes requested):
Thanks for all the changes. This is definitely moving in the right direction. I hope the new comments help to push this even closer to the finish line.

src/mqt/bench/targets/devices/ibm.py
```diff
@@ -18,7 +18,7 @@
 
 from qiskit.providers.fake_provider import GenericBackendV2
 
-from ..gatesets import get_ibm_eagle_gateset, get_ibm_falcon_gateset, get_ibm_heron_gateset
+from mqt.bench.targets.gatesets.ibm import get_ibm_eagle_gateset, get_ibm_falcon_gateset, get_ibm_heron_gateset
```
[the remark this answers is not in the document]
nquetschlich: Adjusted it accordingly.

tests/devices/test_ionq_device_support.py
```diff
@@ -13,33 +13,32 @@
 import pytest
 from qiskit.transpiler import Target
 
-from mqt.bench.targets.devices.ionq import get_ionq_target
+from mqt.bench.targets.devices.ionq import _build_ionq_target, get_ionq_target  # noqa: PLC2701
```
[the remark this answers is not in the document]
nquetschlich: I removed the test.

src/mqt/bench/targets/gatesets/rigetti.py
nquetschlich: I thinks so, also considering this reference implementation (https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.iSwapGate). Although it is the other way round there. Anyways, I removed this equivalence because it is not necessary.

src/mqt/bench/targets/gatesets/rigetti.py
nquetschlich: I added one for that, although it still has 5 gates (but at least without parameters - not sure how much it improves though).

src/mqt/bench/targets/gatesets/ionq.py
nquetschlich: I followed your hint. Additionally, I add the library now only immediately before the `transpile` method is called. I played around a bit with the `EquivalenceLibrary` and you can get all available decompositions for a gate but would need to compare the existing ones with the one that might be added to check if it is already in there - which is also not ideal.

src/mqt/bench/targets/gatesets/rigetti.py
```diff
@@ -0,0 +1,157 @@
...
+    qr = QuantumRegister(1, "q")
+    qc = QuantumCircuit(qr)
+
+    qc.rz(-np.pi / 2, qr[0])
+    qc.append(RXPI2Gate(), [qr[0]])
+    qc.rz(theta, qr[0])
+    qc.append(RXPI2DgGate(), [qr[0]])
+    qc.rz(np.pi / 2, qr[0])
+
+    sel.add_equivalence(RXGate(theta), qc)
```
burgholzer, line 151: This doesn't really help much I suppose.
What I actually meant with my other comment was to add equivalences from 
`RXGate(pi/2)` to `RXPI2Gate` and so on.

burgholzer (changes requested):
One final comment, then I believe this should be good to go here 🙌

src/mqt/bench/targets/gatesets/rigetti.py
nquetschlich: Is this what you would like to see here?
```python
def rx_gate_equivalences(sel: EquivalenceLibrary) -> None:
    """Add RX(θ) gate equivalence using RX(±π/2) and RZ gates."""
    qr = QuantumRegister(1, "q")
    qc = QuantumCircuit(qr)
    qc.append(RXPI2Gate(), [qr[0]])
    sel.add_equivalence(RXGate(np.pi / 2), qc)

    qr = QuantumRegister(1, "q")
    qc = QuantumCircuit(qr)
    qc.append(RXPI2DgGate(), [qr[0]])
    sel.add_equivalence(RXGate(-np.pi / 2), qc)

    qr = QuantumRegister(1, "q")
    qc = QuantumCircuit(qr)
    qc.append(RXPIGate(), [qr[0]])
    sel.add_equivalence(RXGate(np.pi), qc)
```

If yes: Apparently it is not possible to add these equivalences only for specific angles.

burgholzer: Yeah. Something like that is what I would have expected.
How's that not supported? Or what is the error you are seeing?

nquetschlich: It leads to an internal Qiskit Error in `qiskit.transpiler.passes.basis.basis_translator.BasisTranslator.run`: `AttributeError: 'float' object has no attribute '_uuid'` because for a parameterized gate, Qiskit apparently expects a parameter instead of just a specific angle.

burgholzer: Hm.. unfortunate..
That's something that I would consider opening up an issue on the Qiskit tracker for.
Such equivalences could be extremely helpful.

Any how.. I suppose this PR can be merged then 😌

burgholzer (approved):

PR #571, opened 2025-05-23
:sparkles: Distinct `get_benchmark` function for all levels

simon1hofmann:


src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +153,217 @@ def get_native_gates_level(
...
-        "vqetwolocalrandom",
-        "wstate",
-        "shor",
-    ]
+def get_mapped_benchmark(
+    benchmark: str | QuantumCircuit,
+    circuit_size: int | None = None,
+    benchmark_instance_name: str | None = None,
+    opt_level: int | None = 2,
+    target: Target = None,
```
nquetschlich, line 282: We should make this parameter mandatory.

src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +153,217 @@ def get_native_gates_level(
...
-        "vqesu2random",
-        "vqetwolocalrandom",
-        "wstate",
-        "shor",
-    ]
+def get_mapped_benchmark(
+    benchmark: str | QuantumCircuit,
+    circuit_size: int | None = None,
+    benchmark_instance_name: str | None = None,
+    opt_level: int | None = 2,
```
nquetschlich, line 281: I would also make this mandatory but default to `2`.

src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +153,217 @@ def get_native_gates_level(
...
-        to_be_factored_number, a_value = lib.get_instance(benchmark_instance_name)
-        qc = lib.create_circuit(to_be_factored_number, a_value)
 
-    else:
-        qc = lib.create_circuit(circuit_size)
+def get_native_gates_benchmark(
+    benchmark: str | QuantumCircuit,
+    circuit_size: int | None = None,
+    benchmark_instance_name: str | None = None,
+    opt_level: int | None = 2,
```
nquetschlich, line 252: I would also make this mandatory but default to 2.

src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +153,217 @@ def get_native_gates_level(
...
+    target: Target = None,
```
nquetschlich, line 253: I would also make this mandatory without a default value.

src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +153,217 @@ def get_native_gates_level(
...
+        level: Choice of level, either as a string ("alg", "indep", "nativegates" or "mapped") or as a number between 0-3 where 0 corresponds to "alg" level and 3 to "mapped" level
+        circuit_size: Input for the benchmark creation, in most cases this is equal to the qubit number
+        benchmark_instance_name: Input selection for some benchmarks, namely "shor"
+        opt_level: Optimization level to be used by the transpiler.
+        target: `~qiskit.transpiler.target.Target` for the benchmark generation (only used for "nativegates" and "mapped" level)
+
+    Returns:
+        Qiskit::QuantumCircuit object representing the benchmark with the selected options
+    """
+    level_map: dict[str | int, str] = {
```
nquetschlich, line 327: I think we should actually move away from having both the `str` *and* the `int` description of the level. I would prefer to only have the `str` one, using an enum would probably even nicer. This should reduce the complexity and different possible cases to make the code easier.

nquetschlich (changes requested):
Thanks a lot, I really like all the simplifications and the refactoring introduced here. Just added a few comments.

One large concern is the addition of the internal `_to_mapped` as well as the ones for the other levels. As far as I can see, they are only called once in the respective `get_<level>_benchmark` method. Therefore, I would prefer having just a single combined method for each level. But feel free to disagree, maybe I am overlooking something.

src/mqt/bench/benchmark_generation.py
```diff
@@ -10,275 +10,196 @@
...
+        "quarkcardinality",
+        "quarkcopula",
+        "qwalk",
+        "randomcircuit",
+        "shor",
+        "vqerealamprandom",
+        "vqesu2random",
+        "vqetwolocalrandom",
+        "wstate",
+    ]
```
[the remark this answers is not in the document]
simon1hofmann: Sure, I already talked with @nquetschlich about that but decided to move that to a separate PR.

src/mqt/bench/benchmark_generation.py
```diff
@@ -10,275 +10,196 @@
...
-    opt_level: int,
-    file_precheck: bool,
-    return_qc: bool = False,
-    target_directory: str = "./",
-    target_filename: str = "",
-    output_format: OutputFormat = OutputFormat.QASM3,
-) -> bool | QuantumCircuit:
-    """Handles the creation of the benchmark on the target-dependent native gates level.
+    circuit_size: int | None = None,
+    benchmark_instance_name: str | None = None,
```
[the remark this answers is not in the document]
simon1hofmann: But then there is still the optional `circuit_size` argument.

src/mqt/bench/benchmark_generation.py
```diff
@@ -10,286 +10,206 @@
...
-@dataclass
-class QiskitSettings:
-    """Data class for the Qiskit compiler settings."""
+# qubits → (nickname, N, a)
+SHOR_SIZE_TO_PARAMS = {
+    18: ("small", 15, 4),
+    42: ("medium", 821, 4),
+    58: ("large", 11777, 4),
+    74: ("xlarge", 201209, 4),
+}
```
burgholzer, line 46: Hm. I actually had in mind to put all of this Shor-specific code into the Shor.py file and have it expose the same function as all the other benchmarks that just takes the circuit size.
That would isolate this special handling to the class where it is necessary and would decouple it from the benchmark generation itself, which would become more uniform.
It could be noted in the documentation of the Shor class, that this algorithm is only limited for certain numbers of qubits.

src/mqt/bench/benchmark_generation.py
```diff
@@ -300,225 +197,96 @@ def get_native_gates_level(
...
-        to_be_factored_number, a_value = lib.get_instance(benchmark_instance_name)
-        qc = lib.create_circuit(to_be_factored_number, a_value)
-
-    else:
-        qc = lib.create_circuit(circuit_size)
-
-    if level in ("alg", 0):
-        return qc
+    if level is BenchmarkLevel.ALG:
+        print(type(benchmark))
```
burgholzer, line 258: This seems like a debugging leftover?
```suggestion
```

burgholzer (changes requested):
After a good night of sleep, I have a couple of additional thoughts on this.
I'd hope that this is the final round of iteration 😌

src/mqt/bench/benchmark_generation.py
```diff
@@ -10,286 +10,183 @@
...
-    opt_level: int,
-    file_precheck: bool,
-    return_qc: bool = False,
-    target_directory: str = "./",
-    target_filename: str = "",
-    output_format: OutputFormat = OutputFormat.QASM3,
-) -> bool | QuantumCircuit:
-    """Handles the creation of the benchmark on the target-independent level.
+def get_benchmark_alg(
+    benchmark: str | QuantumCircuit,
```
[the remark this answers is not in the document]
burgholzer: Yeah. That also seems like a reasonable solution! 👍🏻
Seems like a good compromise.

burgholzer (approved):
Yeah, this looks great now! Thanks for all the iterations. Really happy with the result. 🙌

PR #572, opened 2025-05-23
:art: Refactor hard-coded `get_available_devices()` function

simon1hofmann:


burgholzer (commented):
This is looking great. I very much like the idea. Just two thoughts:
- this should probably be rebased on top of the other open PR that changes some of the devices and gatesets
- as Nils pointed out, it would be great to handle gatesets in a similar fashion.

src/mqt/bench/targets/devices/__init__.py
```diff
@@ -13,51 +13,37 @@
 from functools import cache
 from typing import TYPE_CHECKING
 
-from .ibm import get_ibm_target
-from .ionq import get_ionq_target
-from .iqm import get_iqm_target
-from .quantinuum import get_quantinuum_target
-from .rigetti import get_rigetti_target
+from . import ibm, ionq, iqm, quantinuum, rigetti  # noqa: F401
```
nquetschlich, line 16: Why is this warning filter necessary?

src/mqt/bench/targets/gatesets/rigetti.py
```diff
@@ -15,7 +15,10 @@
 from qiskit.circuit import EquivalenceLibrary, Gate, Parameter
 from qiskit.circuit.library.standard_gates import RXGate, RZGate, UGate
 
+from ._registry import register
 
+
+@register("rigetti")
```
nquetschlich, line 21: Either we should also name the gateset to `rigetti_ankaa` or adjust the method name below to `get_rigetti_gateset`. I would the latter.

tests/test_devices.py
```diff
@@ -0,0 +1,213 @@
+# Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+# Copyright (c) 2025 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""Test the IBM devices."""
```
nquetschlich, line 9: The docstring is not correct here. Also, I suggest to rename the file to `test_targets` since also the native gatesets are validated here.

Some further thoughts:
- Should we cover all available devices here? Right now, this is the case for IBM but not for IonQ.
- Can we refactor the tests to avoid code duplication? One problem might be that some devices provide additional duration data.
- The test function names should be streamlined as well. Right now, some are still mentioning `from_calibration`.

nquetschlich (changes requested):
Thanks a lot for this PR, @simon1hofmann. This makes the target handling much better. 

I just added a few minor comments that are easy to address. Before merging, I assume that @burgholzer also likes to have a look.

src/mqt/bench/targets/gatesets/rigetti.py
simon1hofmann: I renamed it to `get_rigetti_gateset`.

simon1hofmann:
> I can have a look later on, once Simon is through with your requested changes.

@burgholzer I included Nils' feedback if you want to review this PR now.

tests/test_targets.py
```diff
@@ -0,0 +1,216 @@
...
+    from collections.abc import Mapping, Sequence
+
+
+@dataclass(frozen=True)
+class DeviceSpec:
+    """Specification describing an expected device configuration."""
+
+    name: str
+    num_qubits: int
+    single_gates: set[str] = field(default_factory=set)
```
burgholzer, line 32:
```suggestion
    single_qubit_gates: set[str] = field(default_factory=set)
```
Just for consistency.

tests/test_targets.py
burgholzer, line 1: I only have one request for an additional test: defining and registering an additional target in the tests and making sure it is available from the registry. 
Actually, a similar test for a test gateset.

Other than that. The tests here are great 🙏

burgholzer (changes requested):
This looks great! 🙏
Just two fairly small requests regarding testing.

tests/test_targets.py
simon1hofmann: Thanks a lot for the feedback, I added more tests and also streamlined the implementation of `get_available_devices` and `get_available_gatesets`.

burgholzer (approved):
Really great to see this! Thanks for all the changes 🙏

burgholzer:
Now it's up to @nquetschlich to approve and merge 😉

nquetschlich (approved):

PR #577, opened 2025-05-24
✨ feat: implement mirror circuit generation for device benchmarking

### LINKED ISSUE #439: ✨ Option to generate mirror circuits for reliable benchmarking of devices

When benchmarking the performance of quantum devices, it is always desirable to have a deterministic outcome of a benchmark that is efficient to verify.
While this is possible for algorithms such as Grover (has the desired entry been found), or Shor (does the result allows to factorize the integer), it is much harder in practice for all kinds of variational algorithms as these typically have a very continuous distribution of measurement outcomes that is hard to predict and verify.
Mirror circuits (https://arxiv.org/abs/2112.09853) have been proposed to address this.
They take a circuit and join it with its inverse so that any execution of the circuit is expected to yield the all-zero state $\ket{0\dots 0}$ with certainty.

MQT Bench should offer the option to generate mirror circuits.
Constructing these circuits is fairly easy in code, so this shouldn't be too big of a deal.
Given how such circuits will be predominantly used for evaluating hardware, I believe it makes sense to only offer this option on the mapped level.

One should be careful that the resulting circuits are still valid and executable on the targeted device. One should also avoid to apply any further optimizations on the resulting mirror circuit as these could lead to the circuit collapsing to the empty circuit, which would defeat the whole purpose of constructing them.

CreativeBinBag:
Closes #439

This PR adds functionality to generate "mirror circuits" (Circuit @ Circuit.inverse()) to MQT Bench. 

**Implementation**
*   A `generate_mirror_circuit: bool` parameter is added to `get_benchmark` and individual level-specific benchmark functions.
*   When `True`, an internal helper `_create_mirror_circuit` constructs the mirror circuit by:
    1.  Taking the base circuit (U).
    2.  Composing U with its inverse (U_inv), separated by a barrier.
    3.  Adding appropriate measurements at the end of the combined circuit.
*   For `mapped` level circuits, the `final_layout` of the mirror circuit is set to a trivial layout
*   The CLI is updated with a `--mirror` flag.


nquetschlich (dismissed):
Hi @CreativeBinBag, thank you very much for your contribution. This is highly appreciated and we are happy to see your interest in adding a mirror circuit functionality to MQT Bench. 

I really like the direction this PR is heading and would like to fine-tune it a bit more to achieve a tighter integration. To this end, I added some comments. The most important point is to make the mirror functionality available on all levels by introducing a parameter to the `get_benchmark` method. There, you could check for the return statement of all levels whether the `mirror_circuit` parameter is set to True and, if so, return a mirror circuit instead. To this end, a `mirror_circuit` functionality should be defined in its own function that applies these changes. 

Please feel free to discuss any of the mentioned aspects and always reach out in case of questions. 

Looking forward to have this functionality integrated in MQT Bench :)

CreativeBinBag:
Hi @nquetschlich, thank you so much for the detailed review and the insightful comments! I really appreciate you taking the time. I'll go ahead and incorporate your suggestions, but I wanted to discuss a couple of your points to make sure I'm on the right track:

> In general, we try to minimize the necessary filtered warnings. Is this one really necessary here? I think we should be able to remove it.

I added that filter because when I was running the tests locally, I kept seeing a DeprecationWarning related to qiskit.providers.models (specifically, "qiskit.providers.models is deprecated since Qiskit 1.2"). My aim was just to keep the test output clean.
I'm definitely on board with minimizing filtered warnings. Do you have any initial thoughts on what part of the Qiskit usage might be triggering this, or if there's a known, straightforward way to update the code to avoid this deprecated module?

> I think adding the mirror circuit as a separate level might exclude certain use cases of users who might like to have mirrored circuit on all abstraction levels. Therefore, I rather would prefer a solution with a optional parameter in the get_benchmark function that is something like mirror_circuit: bool = False to enable this instead of defining a separate level.

I completely see your point that having a mirror_circuit flag could be more flexible and allow mirroring at other levels (like alg or indep), which could be useful for different kinds of analyses. My thinking for creating a distinct "mirror" level was to directly address the specific request in Issue #439 by @burgholzer. The issue mentioned: "Given such circuits will be predominantly used for evaluating hardware. I believe it makes sense to only offer this option on the mapped level." So the current implementation focuses on providing this C C† structure specifically for mapped circuits, as that seemed to be the primary use case for hardware verification. Let me know if you'd still like me to go with the mirror_circuit flag approach.

> I think the try and except statements are not necessary since the chosen device is available by default.

My reasoning for including the try-except ValueError: pytest.skip(...) around loading ibm_falcon_27 was just to make the test a bit more robust in case it's run in an environment where that specific device data might not be perfectly set up or available, which could cause get_device_by_name to raise a ValueError.
If ibm_falcon_27 (or its data) is indeed guaranteed to be available and loadable in all the testing environments then you're absolutely right,  and the try-except isn't strictly needed. I'd be happy to remove it if you confirm that's the case. I just erred on the side of caution there.

Thanks again for your guidance! Let me know your thoughts on these points.

burgholzer:
Hey @CreativeBinBag 👋🏻

Many thanks for your contribution here! Much appreciated.

I just wanted to quickly chip in my opinion on the separate level vs. additional flag discussion.
As I wrote in the original issue, I believe that it only makes sense to create these circuits from the mapped level circuits, where the circuits are not transformed further.
A mirror circuit on any other level would always be optimized away completely by any reasonably sophisticated compiler.
I personally do not see much value in offering mirror circuits on these other levels and feel like the extra level solution is quite clean.

Another small side comment: are you still seeing the deprecation warning when you use the latest Qiskit version? These should be gone with Qiskit 2.0 I believe.

nquetschlich:
Thanks for both your responses.

I understand the argumentation why the focus on the mapped level makes sense. However, I am wondering whether there are other use cases on higher levels, e.g., evaluating the correctness of a (noisy) simulator. Since the logic to generate mirror circuits would be the same of all levels, I do not see any harm in offering such circuits since it does not cause any additional effort and might address further use cases. However, I do not have a very strong opinion here and see the more complex interface with an additional parameter as a minor disadvantage of this solution.

Regarding the discussion of an additional parameter vs an own level: Here, I do have a strong opinion and think that a parameter (either for all levels or just for the mapped one, depending on the outcome of the discussion above) is advantageous. In general, our current benchmark levels address different abstraction levels. Since the mirror circuit does not change these abstractions, I clearly see the connection to the respective level, e.g., the mapped one - it "just" adds more gates of the same level and, in my opinion, this does not justify a new level. Furthermore, in the current implementation, `get_mapped_level` is called, which clearly shows the direct connection to the considered level (mapped in this case). Therefore, I find it more intuitive to have the implementation closer to the implementation of that level.

To me, something like a pulse-level representation that provides the circuit in a format even closer to the hardware/execution would justify a new, own level.

nquetschlich:
> 
> > I think the try and except statements are not necessary since the chosen device is available by default.
> 
> My reasoning for including the try-except ValueError: pytest.skip(...) around loading ibm_falcon_27 was just to make the test a bit more robust in case it's run in an environment where that specific device data might not be perfectly set up or available, which could cause get_device_by_name to raise a ValueError. If ibm_falcon_27 (or its data) is indeed guaranteed to be available and loadable in all the testing environments then you're absolutely right, and the try-except isn't strictly needed. I'd be happy to remove it if you confirm that's the case. I just erred on the side of caution there.
> 
> Thanks again for your guidance! Let me know your thoughts on these points.

Coming back to this comment, I think it is a fair assumption that the device is available. While I agree with your argumentation, many of the other tests are based on the assumption that certain devices are available and do not have this error handling. Therefore, I'd prefer to keep it consistent. In the case of a missing but expected device, we need to update the respective tests anyways.

burgholzer:
> I understand the argumentation why the focus on the mapped level makes sense. However, I am wondering whether there are other use cases on higher levels, e.g., evaluating the correctness of a (noisy) simulator. Since the logic to generate mirror circuits would be the same of all levels, I do not see any harm in offering such circuits since it does not cause any additional effort and might address further use cases.

I agree with that sentiment. It's not really doing any harm to add the option for further levels.

> Regarding the discussion of an additional parameter vs an own level: Here, I do have a strong opinion and think that a parameter (either for all levels or just for the mapped one, depending on the outcome of the discussion above) is advantageous. In general, our current benchmark levels address different abstraction levels. Since the mirror circuit does not change these abstractions, I clearly see the connection to the respective level, e.g., the mapped one - it "just" adds more gates of the same level and, in my opinion, this does not justify a new level. Furthermore, in the current implementation, `get_mapped_level` is called, which clearly shows the direct connection to the considered level (mapped in this case). Therefore, I find it more intuitive to have the implementation closer to the implementation of that level.

I agree with that. In principle, the mirror functionality on all levels could work like it is for the mapped level, i.e., calling the respective function for the level, handling measurements, inserting a barrier, inverting the circuit and concatenating them. I am fairly sure this logic should work on all levels.

> To me, something like a pulse-level representation that provides the circuit in a format even closer to the hardware/execution would justify a new, own level.

Agreed 👍🏻

CreativeBinBag:
Hi @nquetschlich and @burgholzer!

Once again thank you very much for your helpful feedback! I've pushed an update incorporating your suggestions. My implementation now uses an optional generate_mirror_circuit boolean parameter in get_benchmark and a corresponding --mirror flag in the CLI. I removed the unnecessary .copy() before measure_all(). Also added a test to verify that the unitary part of the generated mirror circuit indeed collapses to an identity operation.

Regarding the try-except blocks in test_get_benchmark_mirror_option, the initial device/target loading is now direct, assuming availability as suggested. The try-except blocks that remain are specifically for the unitary verification steps to handle potential runtime issues like memory constraints or unexpected transpiler behavior during those complex checks. 

@burgholzer Regarding the Qiskit deprecation warning: Yes, I am still encountering DeprecationWarnings even with the latest Qiskit version. I'm currently using pytest -W "ignore::DeprecationWarning" tests/ to suppress them during local testing.

Please let me know if there's anything else that needs to be changed. Thank you!

src/mqt/bench/benchmark_generation.py
```diff
@@ -55,30 +55,33 @@ def generate_filename(
     benchmark_name: str,
     level: str,
     num_qubits: int | None,
+    is_mirror: bool = False,  # New parameter
```
burgholzer, line 58: This should probably be named `generate_mirror_circuit`
And it should probably be the last parameter in the argument list.

src/mqt/bench/benchmark_generation.py
```diff
@@ -55,30 +55,33 @@ def generate_filename(
...
-        opt_level: optional optimization level (used for 'nativegates' and 'mapped')
+        opt_level: optional optimization level (used for 'indep', 'nativegates', and 'mapped')
 
     Returns:
         A string representing a filename (excluding extension) that encodes
         all relevant metadata for reproducibility and clarity.
     """
-    base = f"{benchmark_name}_{level}"
+    level_str_modified = f"{level}_mirror" if is_mirror else level
+    base = f"{benchmark_name}_{level_str_modified}"
```
burgholzer, line 77: Would have probably written this as something like (untested)
```suggestion
    base = f"{benchmark_name}_{level}{'_mirror' if generate_mirror_circuit}"
```
Highly likely that the above is not yet syntactically correct, but you get the idea

src/mqt/bench/benchmark_generation.py
```diff
@@ -443,50 +452,62 @@ def get_benchmark(
...
+
+    if generate_mirror_circuit:
+        # Get the unitary part of the processed circuit (remove measurements)
+        qc_unitary_base = processed_qc.remove_final_measurements(inplace=False)
+
+        # Compute its inverse
+        qc_inverse = qc_unitary_base.inverse()
+
+        # Compose them. The circuit qc_unitary_base is modified by compose.
+        mirrored_qc_unitary = qc_unitary_base.compose(qc_inverse)
```
burgholzer, line 501: Circuit can be composed inplace to avoid another extra copy of the circuit.

src/mqt/bench/benchmark_generation.py
```diff
@@ -443,50 +452,62 @@ def get_benchmark(
...
+            raise RuntimeError(msg)  # pragma: no cover
+
+    if generate_mirror_circuit:
+        # Get the unitary part of the processed circuit (remove measurements)
+        qc_unitary_base = processed_qc.remove_final_measurements(inplace=False)
+
+        # Compute its inverse
+        qc_inverse = qc_unitary_base.inverse()
+
+        # Compose them. The circuit qc_unitary_base is modified by compose.
```
burgholzer, line 500: I am fairly sure we want to insert a global barrier after the first circuit before concatenation do delimit the circuits and to try to prevent optimizers from optimizing away the mirror circuit.

src/mqt/bench/benchmark_generation.py
```diff
@@ -443,50 +452,62 @@ def get_benchmark(
...
+        mirrored_qc_unitary = qc_unitary_base.compose(qc_inverse)
+        # Set a more descriptive name based on the processed circuit's name.
+        mirrored_qc_unitary.name = f"{processed_qc.name}_mirror"
+
+        # Add measurements back to the final mirrored circuit.
+        final_mirrored_qc = mirrored_qc_unitary
```
burgholzer, line 506: This assignment is unnecessary. Just continue operating on the circuit you have built above.
In the worst case, this might create an extra circuit copy.

tests/test_bench.py
```diff
@@ -836,3 +869,113 @@ def test_custom_target() -> None:
...
+        if qc_mirror_unitary.num_qubits <= max_qubits_for_operator_check:
+            identity_op_circuit = QuantumCircuit(qc_mirror_unitary.num_qubits)
+            identity_op = Operator(identity_op_circuit)
+            try:
+                mirror_op = Operator(qc_mirror_unitary)
+                assert mirror_op.equiv(identity_op), (
+                    f"Unitary part of mirror circuit (level '{level_str}', {qc_mirror_unitary.num_qubits} qubits) "
+                    f"is not equivalent to identity. QASM: {qc_mirror_unitary.qasm()}"
+                )
+            except Exception as e:  # Catch potential memory errors even for small circuits if they are dense
```
burgholzer, line 972: This try and except is certainly not necessary.
The operator of a 4 qubit circuit is a 16*16 matrix, which takes 128B per entry. That makes 32KiB of memory.

burgholzer:
Also, you can simply apply the suggestions from the review in a batch, by going to the files changed tab on GitHub and clicking the "add suggestion to batch" button.
This will automatically resolve the respective comments.

CreativeBinBag:
Hi @burgholzer 👋🏻
Thank you very much for being so patient and thoroughly reviewing my PR!
I've pushed updates addressing your feedback. The PR description reflects these changes. Could you please re-review? Thanks!

tests/test_bench.py
```diff
@@ -797,3 +842,109 @@ def test_assert_never_runtime() -> None:
...
+            benchmark=benchmark_name,
+            level=level_enum,
+            circuit_size=logical_circuit_size,
+            opt_level=comp_opt_level if comp_opt_level is not None else 2,
+            target=target_obj,
+            generate_mirror_circuit=True,
+        )
+
+        assert qc_mirror is not None
+        assert isinstance(qc_mirror, QuantumCircuit)
```
burgholzer, line 890:
```suggestion
```
Already guaranteed by static typing

tests/test_bench.py
```diff
@@ -797,3 +842,109 @@ def test_assert_never_runtime() -> None:
...
+        qc_mirror = get_benchmark(
+            benchmark=benchmark_name,
+            level=level_enum,
+            circuit_size=logical_circuit_size,
+            opt_level=comp_opt_level if comp_opt_level is not None else 2,
+            target=target_obj,
+            generate_mirror_circuit=True,
+        )
+
+        assert qc_mirror is not None
```
burgholzer, line 889:
```suggestion
```
There is no path through the get_benchmark function that would return None IIRC.

tests/test_bench.py
```diff
@@ -797,3 +842,109 @@ def test_assert_never_runtime() -> None:
...
+            assert qc_mirror.num_qubits == qc_base.num_qubits
+        else:
+            assert qc_mirror.num_qubits == logical_circuit_size
+
+        assert qc_mirror.num_clbits == qc_base.num_clbits
+
+        if qc_base.num_clbits > 0:
+            assert any(inst.operation.name == "measure" for inst in qc_mirror.data)
+        else:
+            assert not any(inst.operation.name == "measure" for inst in qc_mirror.data)
```
burgholzer, line 903: This does not make any sense to me. Every circuit must have measurements, so it must have classical bits.

tests/test_bench.py
```diff
@@ -797,3 +842,109 @@ def test_assert_never_runtime() -> None:
...
+        )
+
+        # --- Layout Verification---
+        if level_enum == BenchmarkLevel.MAPPED and target_obj:
+            assert isinstance(qc_base.layout, TranspileLayout), (
+                f"Base mapped circuit for {benchmark_name} lacks TranspileLayout."
+            )
+            assert isinstance(qc_mirror.layout, TranspileLayout), (
+                f"Mirror of mapped circuit for {benchmark_name} lacks TranspileLayout."
+            )
```
burgholzer, line 916: You can simply check if the layout is None or not. The type will always be TranspileLayout if it is not None.

tests/test_bench.py
```diff
@@ -797,3 +842,109 @@ def test_assert_never_runtime() -> None:
...
+        )
+
+        all_ops_are_identity_or_empty = all(instr.operation.name == "id" for instr in optimized_circuit.data)
+        assert all_ops_are_identity_or_empty, (
+            f"Unitary part of mirror (U@U_inv, MQT Bench barrier removed for this check) "
+            f"for level '{level_enum.name}' ({qc_mirror.num_qubits} qubits) "
+            f"did not optimize to only identity gates or an empty circuit. "
+            f"Found operations: {[op.operation.name for op in optimized_circuit.data if op.operation.name != 'id']}. "
+            f"Optimized QASM: {dumps(optimized_circuit)}"
+        )
```
burgholzer, line 950: Just out of interest:
Would removing the Id hate from the gateset completely reduce the circuit to an empty circuit?
I'd very much prefer that.

burgholzer (changes requested):
Thanks for another iteration. We are getting there. Still some things to iron out before this can go in 😌

burgholzer:
@CreativeBinBag may I ask about an update here? Is something blocking your progress?

Note that some PRs have been merged in the meantime and that you will have to resolve the conflicts with the main branch. Shouldn't be too bad.

CreativeBinBag:
@burgholzer Thank you for your patience. Unfortunately, I’ve been down with COVID for over a week, which has made it difficult to work at my screen for long. Give me a day or two and I’ll wrap this up

burgholzer:
> @burgholzer Thank you for your patience. Unfortunately, I’ve been down with COVID for over a week, which has made it difficult to work at my screen for long. Give me a day or two and I’ll wrap this up

Oh no, how unfortunate. Hope you get better soon! 🙏🏼 
Please do not feel rushed. We have all the time in the world for this. So take some rest and come back to this PR when you are feeling well again 😌

CreativeBinBag:
Hi @nquetschlich, just wanted to let you know that I'm almost done with the changes. I’ll push them to this branch soon! Thank you so much for understanding.

CreativeBinBag:
@burgholzer @nquetschlich please review the changes whenever you have time. I’m so sorry once again for the delay in addressing the last feedback comments. The past few weeks have been very rough for me, but I want to assure you that if you have more feedback, I can take care of it ASAP now!

About this comment: 

> If this is handled by Qiskit already (was not aware of that), then all the logic here regarding that is redundant and I'd rather remove it to keep this simple.

the final mirror circuit's final_layout doesn't always end up being identical to the original initial_layout, which is a strict requirement for this benchmark. My explicit code just enforces that one important final step to guarantee the circuit is a true identity on the physical qubits. I've updated the docstring to make this clearer, I think the previous version was pretty vague!

burgholzer:
> @burgholzer @nquetschlich please review the changes whenever you have time. I’m so sorry once again for the delay in addressing the last feedback comments. The past few weeks have been very rough for me, but I want to assure you that if you have more feedback, I can take care of it ASAP now!
> 
> About this comment:
> 
> > If this is handled by Qiskit already (was not aware of that), then all the logic here regarding that is redundant and I'd rather remove it to keep this simple.
> 
> the final mirror circuit's final_layout doesn't always end up being identical to the original initial_layout, which is a strict requirement for this benchmark. My explicit code just enforces that one important final step to guarantee the circuit is a true identity on the physical qubits. I've updated the docstring to make this clearer, I think the previous version was pretty vague!

Hey 👋🏼 
No need to be sorry! Hope you are feeling better again!
Thanks for another iteration here. 
I'll do a small final iteration on the code myself. Would you mind giving me push access to your fork so that I can directly push my changes to this PR?

burgholzer (approved):
Alright. I pushed a couple of changes to the PR that simplify the overall implementation.

These commits also fix the `final_layout` handling. As per the Qiskit docs, the final layout only describes the permutation caused by SWAP insertion during routing. As such it is, per definition, the identity/trivial layout since we are reversing every SWAP as part of the mirroring.

The mirror circuit construction also used one less copy and only measures the active qubits in the end.

Lastly, the Boolean arguments have been turned into keyword-only arguments, which is a best practice to avoid passing random sequences of Booleans to functions.

If CI passes, I am happy with this PR and will go ahead and merge it.
@CreativeBinBag Many thanks for your contribution here. Much appreciated. 🙏🏼 And congrats on the first contribution to the Munich Quantum Toolkit 🎉

PR #581, opened 2025-05-26
✨ add option to parameterized benchmarks to create circuits with symbolic parameters

### LINKED ISSUE #542: ✨ Symbolic Parameter for Variational Quantum Algorithm Benchmarks

Currently, many quantum algorithms that actually have parameterized gates are provided with random parameter values. 

Since now OpenQASM 3 is supported, all such benchmarks (such as VQE, QAOA, etc.) should also be optionally provided with actual symbolic parameters instead of randomly initialized values.

nquetschlich:
So far, always random parameter values were used in benchmarks with actual parameters. Now, it is also possible to create circuits with symbolic parameters instead. This can be triggered using a newly introduced boolean parameter that defaults to `True` to not change the default behavior.

This PR resolves #542.


src/mqt/bench/benchmarks/vqerealamp.py
```diff
@@ -24,20 +25,27 @@
     from qiskit.circuit import QuantumCircuit
 
 
-def create_circuit(num_qubits: int) -> QuantumCircuit:
-    """Returns a quantum circuit implementing the RealAmplitudes ansatz with random parameter values.
+def create_circuit(num_qubits: int, reps: int = 3, random_parameters: bool = True) -> QuantumCircuit:
```
burgholzer, line 28: Is this also the default for the Qiskit function?
I think we should be matching Qiskit's default for all of these algorithms

burgholzer (changes requested):
Thanks for the PR. Definitely moving in the right direction.
I'd take it one step further though, see the comments.
I am pretty convinced that would be a cleaner solution to the overall problem.

nquetschlich:
> Thanks for the PR. Definitely moving in the right direction. I'd take it one step further though, see the comments. I am pretty convinced that would be a cleaner solution to the overall problem.

That is a neat idea. I already adjusted the benchmarks to incorporate all your feedback. As soon as #571 is merged, I will add the random initialization logic to avoid merge conflicts.

src/mqt/bench/benchmarks/vqerealamp.py
nquetschlich: Done.

src/mqt/bench/benchmarks/bmw_quark_copula.py
```diff
@@ -37,41 +36,55 @@ def create_circuit(num_qubits: int, depth: int = 2) -> QuantumCircuit:
...
+    # === Compute number of parameters ===
+    num_single_qubit_gates = depth * n_registers * n * 3
+    num_rxx_gates = depth * n_registers * comb(n, 2)
+    total_params = num_single_qubit_gates + num_rxx_gates
+
+    param_vector = ParameterVector("p", total_params)
+
+    param_index = 0
+
+    def get_param() -> float | Parameter:
```
burgholzer, line 54: The return value of this function will never be float.

src/mqt/bench/benchmarks/qnn.py
```diff
@@ -25,15 +24,13 @@ def create_circuit(num_qubits: int) -> QuantumCircuit:
 
     Arguments:
         num_qubits: number of qubits of the returned quantum circuit
+        random_parameters: If True, assign random parameter values; if False, use symbolic parameters.
```
burgholzer, line 27: Leftover parameter description
```suggestion
```

src/mqt/bench/benchmarks/qaoa.py
```diff
@@ -11,30 +11,26 @@
...
-        seed: Random seed for reproducibility.
 
     Returns:
         QuantumCircuit: Quantum circuit implementing QAOA.
     """
-    # Set the random number generator
-    rng = np.random.default_rng(seed)
-
     # Example adjacency matrix for Max-Cut (toy problem)
+    rng = np.random.default_rng(10)
```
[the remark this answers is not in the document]
nquetschlich: Sure, I just did not see the necessity to have it as a parameter instead of using a fixed seed. What would be the advantage?

src/mqt/bench/benchmarks/bmw_quark_copula.py
nquetschlich: Fixed it.

PR #582, opened 2025-05-26
✨  Implement HHL algorithm

### LINKED ISSUE #216: Re-Implement HHL benchmark
The HHL was removed in #214 because the underlying qiskit functionalities were deprecated. It is desired to re-implement our own version of that benchmark algorithm.

nquetschlich:
Quite some time ago, the HHL implementation based on `qiskit.algorithms` has been removed. Now, the time has come to re-add this benchmark. 

In this current implementation, a simplified version is implemented that solves a specific linear equation system and the `num_qubits` scale the number of QPE qubits. This gives the advantages of getting a circuit that actually has the specified number of qubits in comparison to before. 

However, while I think this implementation could and should be improved in the future, this might be a fair starting point.

This PR resolves #216.

### DIFF
--- a/src/mqt/bench/benchmarks/hhl.py
+++ b/src/mqt/bench/benchmarks/hhl.py
@@ -64,8 +64,6 @@ def create_circuit(num_qubits: int) -> QuantumCircuit:
             inv_lambda = np.clip(inv_lambda, -1, 1)  # valid for arcsin
             theta = 2 * np.arcsin(inv_lambda)
             qc.cry(theta, qr_eig[j], qr_anc[0])
-        else:
-            theta = 0.0
 
     # Step 6: QPE uncomputation (apply QFT + reverse controlled unitary)
     qc.append(QFTGate(num_qpe_qubits), qr_eig)


src/mqt/bench/benchmarks/hhl.py
```diff
@@ -0,0 +1,80 @@
...
+        # Approximate eigenvalue encoded in basis state |j⟩
+        # Use inverse λ (scaled appropriately)
+        estimated_lambda = lam / (2 ** (num_qpe_qubits - j))
+        if estimated_lambda > 0:
+            inv_lambda = 1.0 / estimated_lambda
+            inv_lambda = np.clip(inv_lambda, -1, 1)  # valid for arcsin
+            theta = 2 * np.arcsin(inv_lambda)
+        else:
+            theta = 0.0
+        qc.cry(theta, qr_eig[j], qr_anc[0])
```
burgholzer, line 68: This can be moved to the `if` branch. A rotation around 0 does nothing so we might as well not add it to the circuit.

burgholzer (approved):
This is indeed a fairly limited application of HHL, but still a reasonable one.
In the very end, these circuits are not so different to the QPE circuits we have already. But it is still nice to see.

Random thought: at some point we should actually verify that all of the algorithms here (especially the ones we implement on our own) do the right thing. While not increasingly important from a compiler benchmarking perspective, this becomes increasingly relevant when these algorithms should actually be executed.

PR #731, opened 2025-11-13
✨ Expose seed of `graphstate` generation

ystade:
## Description

Benchmark creation APIs now accept arbitrary configuration options via **kwargs: Unpack[ConfigurationOptions] and propagate them through the pipeline. A new ConfigurationOptions TypedDict with optional seed was added. The graphstate benchmark uses seed for deterministic graph generation; tests for seed reproducibility were added.

## Checklist:

- [ ] I have added migration instructions to the upgrade guide (if needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,7 +89,7 @@ _📚 Refer to the [GitHub Release Notes] for previous changelogs._
 
 <!-- PR links -->
 
-[#731]: https://github.com/munich-quantum-toolkit/bench/pull/709
+[#731]: https://github.com/munich-quantum-toolkit/bench/pull/731
 [#709]: https://github.com/munich-quantum-toolkit/bench/pull/709
 [#705]: https://github.com/munich-quantum-toolkit/bench/pull/705
 [#671]: https://github.com/munich-quantum-toolkit/bench/pull/671


ystade:
@burgholzer pre-commit fails here because I am using the type `Any` to annotate `**kwargs`. What is the correct way to do that?

tests/test_bench.py
```diff
@@ -212,6 +212,35 @@ def test_dj_constant_oracle() -> None:
...
+    #                                 0  1  2  3  4
+    qc_expected = QuantumCircuit(5)
+    qc_expected.h(range(5))
+    qc_expected.cz(0, 1)
+    qc_expected.cz(1, 2)
+    qc_expected.cz(0, 4)
+    qc_expected.cz(2, 3)
+    qc_expected.cz(3, 4)
+    qc_expected.measure_all()
+    qc_actual = get_benchmark_alg("graphstate", 5, seed=42)
```
burgholzer: Wouldn't the better test here be to call the function twice and check that the circuits are the same?

src/mqt/bench/benchmark_generation.py
```diff
@@ -41,6 +43,7 @@ def _get_circuit(
     benchmark: str | QuantumCircuit,
     circuit_size: int | None,
     random_parameters: bool = True,
+    **kwargs: Any,
```
burgholzer: It all depends on what we want here, whether `Any` is just fine here.
If we really want to permit arbitrary kwargs, then `Any` is just fine and a simple `noqa` may be added to all occurrences.
If we only want to permit dedicated, known options, then we might go the route of https://github.com/munich-quantum-toolkit/qcec/blob/549305f2e1705da7eab193a454456ae0ec1dca78/python/mqt/qcec/verify.py#L41 with a typed dictionary that just contains the seed for now)

burgholzer (commented):
Two fairly quick drive-by comments

burgholzer (approved):
This looks great! Many thanks 🙏

PR #806, opened 2026-01-23
Add IQPE benchmark

carlareciofdz:
## Description

Please include a summary of the change and, if applicable, which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes #(issue) 

## Checklist:

- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).

### DIFF
--- a/tests/test_bench.py
+++ b/tests/test_bench.py
@@ -151,8 +151,8 @@ def test_iqpe_exact_structure() -> None:
     Verifies (for a 5-qubit input):
         - Quantum registers: 4 target qubits and 1 ancillary qubit
         - Classical register: num_qubits - 1 (4 bits)
-        - 4 resets and 6 conditional operations
-        - 4 measurements
+        - 8 measurements (4 resets + 4 final measurements)
+        - 10 conditional operations (4 reset conditionals + 6 correction conditionals)
     """
     qc = create_circuit("iqpe_exact", 5)
 
@@ -164,7 +164,7 @@ def test_iqpe_exact_structure() -> None:
     assert ops.get("measure") == 8
 
     if_else_count = sum(1 for inst in qc.data if inst.operation.name == "if_else")
-    assert if_else_count == 10, f"Expected 6 conditional operations, found {if_else_count}"
+    assert if_else_count == 10, f"Expected 10 conditional operations, found {if_else_count}"
 
 
 def test_iqpe_exact_invalid_qubit_number() -> None:


src/mqt/bench/benchmarks/iqpe_exact.py
```diff
@@ -0,0 +1,89 @@
...
+    # We apply an X gate to the q[1] qubit, to prepare the target qubit in the
+    # |1> state
+    qc.x(q[1])
+
+    for k in range(num_bits):
+        index = num_bits - 1 - k
+        # We reset the ancillary qubit in order to reuse in each iteration
+        qc.measure(q[0], c[0])
+        with qc.if_test((c[0], 1)):
+            qc.x(q[0])
```
burgholzer, line 65: Qiskit has a dedicated `reset` instruction that can be used here.
Furthermore, the first reset can be avoided as the qubit will be initialized in `|0>`.

src/mqt/bench/benchmarks/iqpe_exact.py
```diff
@@ -0,0 +1,83 @@
...
+
+    """
+    # We just use 2 quantum registers: q[1] is the main register where
+    # the eigenstate will be encoded, q[0] is the ancillary qubit where the phase
+    # will be encoded. Only the ancillary qubit is measured, and the result will
+    # be stored in "c", the classical register.
+    num_qubits = num_qubits - 1  # because of ancilla qubit
+    q = QuantumRegister(2, "q")
+    c = ClassicalRegister(num_qubits, "c")
+    qc = QuantumCircuit(q, c, name="iqpeexact")
```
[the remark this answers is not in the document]
burgholzer: @carlareciofdz the suggestion by @coderabbitai here is actually wrong. The circuit should only ever have two qubits. The input parameter to the `create_circuit` function only really configures the precision to be used.
I see two ways forward here:
- either we relax the test suite slightly so that it does not assume that the final circuit has `n` qubits.
- or we fix the number of qubits in the `create_circuit` function to `2`; error for everything else; and introduce an additional parameter `precision` to the function that regulates the number of iterations etc.

src/mqt/bench/benchmarks/iqpe_exact.py
```diff
@@ -0,0 +1,83 @@
...
+        for i in range(k):
+            m_index = num_qubits - 1 - i
+            true_angle = -1.0 / (1 << (k - i))
+
+            with qc.if_test((c[m_index], 1)):
+                qc.p(true_angle * np.pi, q[0])
+
+        qc.h(q[0])
+        # We measure and store the result for future corrections
+        qc.measure(q[0], c[k])
```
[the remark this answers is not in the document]
burgholzer: @coderabbitai please check if this is still true

src/mqt/bench/benchmarks/iqpe_exact.py
burgholzer, line 1: This algorithm will probably need a safeguard to limit it to something like 60 qubits because otherwise the shift operations are no longer well defined and will cause issues.
A corresponding check should be added, the docstrings should mention this limitation, and a test is probably also in order.

burgholzer, line 1: This is only covering the `exact` version of iQPE. Is there any particular reason for that?
It could make sense to add one more parameter to the create_circuit function that controls whether the phase is exactly representable or not.
Then, the naming and docstrings can also be adjusted accordingly to refer to this benchmark as IQPE (without the exact or inexact).

tests/test_bench.py
```diff
@@ -145,6 +145,34 @@ def test_adder_circuits(benchmark_name: str, input_value: int, kind: str) -> Non
...
+
+    assert qc.num_qubits == 5
+    assert qc.num_clbits == 4
+
+    ops = qc.count_ops()
+    ops = dict(ops)  # type: dict[str, int]
+    assert ops.get("measure") == 8
+
+    if_else_count = sum(1 for inst in qc.data if inst.operation.name == "if_else")
+    assert if_else_count == 10, f"Expected 10 conditional operations, found {if_else_count}"
```
burgholzer, line 167: This feels a bit hard-coded to be honest. Can we generalize this slightly to cover a few more cases?

flowerthrower:
Dear @carlareciofdz, may I ask for the current status of this PR?

PR #807, opened 2026-01-23
Added Dynamical QFT + Measurement

### LINKED ISSUE #798: ✨ Dynamical QFT
### Problem Statement

The Quantum Fourier Transform is one of the most important subroutines in quantum circuits, particularly in Shor's factoring algorithm. The QFT algorithm, however, uses many 2-qubit gates (O(N^2)) and assumes all-to-all connectivity, which is not applicable to most hardware platforms. To this end, a dynamical circuit version of this algorithm has been developed to resolve some of these issues.

This dynamical circuit version of the QFT immediately followed by measurement (QFT + M) uses O(n) mid-circuit measurements and classical feed-forward instead of O(n^2) 2-qubit gates and removes the assumption of all-to-all connectivity.

**Relevant Literature**
https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.133.150602 

### Proposed Solution

With the growing relevance of dynamical circuits, I think that a dynamical QFT could be a suitable addition to the list of benchmarking circuits.

morrishuismans:
## Description

Adds the Dynamical QFT + Measurement circuit as a dynamical benchmarking circuit.

Fixes #798 

## Checklist:


src/mqt/bench/benchmarks/dynamical_qft.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+# QFT & M v.1
```
burgholzer:
```suggestion
```
The docstring here feels a bit arbitrary

src/mqt/bench/benchmarks/dynamical_qft.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""Dynamical QFT + M benchmark definition."""
```
burgholzer: I would have a slight tendency to consistently refer to this as `Dynamic QFT` (instead of "dynamical" or "QFT+M"). This applies to all occurrences where the name is used.

src/mqt/bench/benchmarks/dynamical_qft.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
+from qiskit.circuit import ClassicalRegister, QuantumCircuit, QuantumRegister
+
+from ._registry import register_benchmark
+
+
+@register_benchmark("dynamical_qft", description="Dynamic Quantum Fourier Transformation and Measurement (DQFT + M)")
+def create_circuit(num_qubits: int) -> QuantumCircuit:
+    """Returns a quantum circuit implementing the Dynamic Quantum Fourier Transform and Measurement algorithm.
```
burgholzer: This docstring could be made a little more detailed and explain a little bit more about the algorithm itself. From this docstring alone, no one would have an idea of what the algorithm is doing. There is also no reference to some kind of resource this is based on.

src/mqt/bench/benchmarks/dynamic_qft.py
```diff
@@ -0,0 +1,40 @@
...
+    q = QuantumRegister(num_qubits, "q")
+    cl = ClassicalRegister(num_qubits, "cl")
+    qc = QuantumCircuit(q, cl, name="dynamical_qft")
+    for i in range(num_qubits):
+        qc.h(q[i])
+        qc.measure(q[i], cl[i])
+        with qc.if_test((cl[i], 1)):
+            for j in range(1, num_qubits - i):
+                qc.p(2 * np.pi / 2 ** (j + 1), q[j + i])
+    return qc
```
burgholzer, line 38: Initially I was thinking this isn't really the way I imagined this to be written, because I would have imagined this to be a single-qubit circuit that repeatedly resets the qubit and applies phase corrections to it based on previous measurements (similar to iQPE from #806).
However, after thinking about it a little more, this kind of does make sense to portray as an `n` qubit circuit. This also allows using this circuit as a benchmark for circuit optimization techniques that would, e.g., introduce qubit reuse to shrink the circuit.

burgholzer: Micro optimization: The phase computation can be simplified a bit here
```
qc.p(2 * np.pi / 2 ** (j + 1), q[j + I])
```
is equivalent to
```
qc.p(np.pi / (2 ** j), q[j + I])
```

burgholzer: Slight naming tweak: The canonical name for classical registers in Qiskit is `c` (not `cl`) might be worth adapting that.

burgholzer: Similar to #806, this will need an upper bound on the number of qubits since the shift operation only makes sense up to a certain number of qubits and is not well defined afterwards.

burgholzer (changes requested):
Thanks for your work on this!
This looks like great progress already.
You can find some suggestions/feedback in the comments below.
I hope none of them should really be hard to address.

You will also need to merge in the latest changes from `main` so that the conflicts are resolved.

src/mqt/bench/benchmarks/dynamic_qft.py
morrishuismans: > Similar to #806, this will need an upper bound on the number of qubits since the shift operation only makes sense up to a certain number of qubits and is not well defined afterwards.

Dear @burgholzer, I am not really sure of what you mean. Do you mean that the Phase gates with the very small phases are unnecessary? If so, would it be better to upper bound the number of qubits or the number of phase gates per layer and discard all gates with negligible phase?
Or do you mean something different with the shift operation?

burgholzer: This is computing `2**j`, which is only well defined up until `j<=63` (or less; needs to be checked) due to integer overflow.
Two solutions basically:
- do not permit generating larger instances
- add an additional check that approximates angles and drops gates if the angle becomes too small

flowerthrower:
Closing due to inactivity. Feel free to reopen if you plan to continue working on it.

PR #816, opened 2026-01-28
Added GHZ dynamic implementation and its tests

### LINKED ISSUE #815: ✨ Dynamic GHZ state generation
### Problem Statement

The GHZ state for n qubits is the superposition state of all qubits being 1 or all qubits being 0.
Example for n=4:
$$\Psi_{GHZ} = \frac{|0000\rangle + |1111\rangle}{\sqrt{2}}$$

**Relevant literature:**
https://arxiv.org/abs/2308.13065
https://arxiv.org/abs/2206.15405

### Proposed Solution

Implement the dynamic GHZ generation circuit using intermediate measurements.

TomasVF:
## Description

Please include a summary of the change and, if applicable, which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes #815

## Checklist:

### DIFF
--- a/tests/test_bench.py
+++ b/tests/test_bench.py
@@ -34,6 +34,7 @@ from qiskit.transpiler import (
 from qiskit.transpiler.passes import GatesInBasis, RemoveBarriers
 
 if TYPE_CHECKING:  # pragma: no cover
+    from collections import OrderedDict
     from collections.abc import Callable
 
     from mqt.bench.configuration_options import ConfigurationOptions
@@ -261,8 +262,9 @@ def test_dynamic_ghz_circuit_structure(num_qubits: int) -> None:
     assert measure_all_regs[0].size == num_qubits
 
     # Check Gate Counts
-    ops = qc.count_ops()  # type: dict[str, int]
-    ops = dict(ops)
+    # TODO: The typing of `QuantumCircuit.count_ops()` is incorrect.
+    #  This will be fixed in Qiskit 2.3.1
+    ops: OrderedDict[str, int] = qc.count_ops()  # ty: ignore[invalid-assignment]
 
     # Check H gates
     expected_h = (num_qubits + 1) // 2


src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+@register_benchmark("ghz_dynamic", description="Dynamic GHZ State")
+def create_circuit(num_qubits: int) -> QuantumCircuit:
+    """Returns a dynamic quantum circuit implementing the GHZ state.
+
+    Arguments:
+        num_qubits: number of qubits of the returned quantum circuit
+    """
+    q = QuantumRegister(num_qubits, "q")
+    c = ClassicalRegister(num_qubits // 2, "mid_measurement")
+    qc = QuantumCircuit(q, c, name="ghz_dynamic")
```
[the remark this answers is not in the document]
TomasVF: It won't fail, Qiskit does allow for ClassicalRegister() to have a value of 0 as long as the quantum register has some elements.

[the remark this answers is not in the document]
TomasVF: And all the uses of c are conditioned on num_qubits beeing greater than 0

tests/test_bench.py
```diff
@@ -224,6 +224,41 @@ def test_graphstate_seed() -> None:
     assert qc_no_seed.name == "graphstate"
 
 
+# Test the dynamic GHZ circuit
+@pytest.mark.parametrize("num_qubits", [1, 2, 3, 7, 10])
+def test_dynamic_ghz_circuit_structure(num_qubits) -> None:
+    # Generate the circuit
+    qc = create_circuit("ghz_dynamic", num_qubits)
```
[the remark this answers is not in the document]
TomasVF: As mentioned in a previous comment Qiskit allows the creation of empty ClassicalRegister() as long as the number of qubits is not 0

TomasVF:
https://github.com/coderabbitai full review

TomasVF:
@coderabbitai full review

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+
+from qiskit.circuit import ClassicalRegister, QuantumCircuit, QuantumRegister
+from qiskit.circuit.classical import expr
+
+from ._registry import register_benchmark
+
+
+@register_benchmark("ghz_dynamic", description="Dynamic GHZ State")
+def create_circuit(num_qubits: int) -> QuantumCircuit:
+    """Returns a dynamic quantum circuit implementing the GHZ state.
```
burgholzer: This would benefit from a little bit more documentation on what the algorithm actually does.
A link to the main reference this is based on would probably also be helpful.

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+        j = i - 1
+        z = i + 1
+        if j >= 0:
+            qc.cx(i, j)
+        if z < num_qubits - 1:
+            qc.cx(i, z)
+
+    j = 0
+
+    # Intermediate measurements on the odd qubits, the if_test statement is there to simulate a reset operation as this is not accepted by some hardware
```
burgholzer, line 53: You may assume that `reset` is supported natively and use it here.
Some qubit technologies have faster reset operations than the feed forward used here so I think it would be good to generally use `reset` here.
I believe this could then be combined with the subsequent loop so that it really becomes a single loop that measures the qubit, computes the condition, applies the correction, resets the right qubits.
At least in my mind that sounds like a very clear structure.

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+    for i in range(1, num_qubits, 2):
+        qc.measure(i, j)
+        with qc.if_test((c[j], 1)):
+            qc.x(i)
+        j += 1
```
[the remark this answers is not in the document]
burgholzer: @TomasVF I have to agree with @coderabbitai here. The code is fairly hard to follow as-is. Could you please use more descriptive names throughout?

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+
+@register_benchmark("ghz_dynamic", description="Dynamic GHZ State")
+def create_circuit(num_qubits: int) -> QuantumCircuit:
+    """Returns a dynamic quantum circuit implementing the GHZ state.
+
+    Arguments:
+        num_qubits: number of qubits of the returned quantum circuit
+    """
+    q = QuantumRegister(num_qubits, "q")
+    c = ClassicalRegister(num_qubits // 2, "mid_measurement")
```
burgholzer: The circuit should have two classical registers: one that is used for the intermediate measurements (that should probably not be called `c`) and a "final" classical register that the `n` qubits are being measured into.
I would want to feed the resulting circuit into a simulator and then evaluate whether the n-qubit register `c` mirrors a GHZ state (50% all-zeroes, 50% all-ones).
Instead of implicitly adding this register in the end via `measure_all`, I would prefer to make this explicit here.

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+        qc.measure(i, j)
+        with qc.if_test((c[j], 1)):
+            qc.x(i)
+        j += 1
+
+    j = 0
+    if num_qubits > 1:
+        condition = c[0]
+
+    # We apply a X gate to all even qubits other than the first one if the exor of the intermediate measurements of all the previous qubits is 1
```
burgholzer:
```suggestion
    # We apply a X gate to all even qubits other than the first one if the XOR of the intermediate measurements of all the previous qubits is 1
```
I agree with the rabbit on this one.

src/mqt/bench/benchmarks/ghz_dynamic.py
```diff
@@ -0,0 +1,70 @@
...
+    for i in range(2, num_qubits, 2):
+        with qc.if_test(expr.equal(condition, True)):
+            qc.x(i)
+        j += 1
+        if j < num_qubits // 2:
```
burgholzer: This can't be quite right, can it? This should apply to every even qubit index until the very end?

src/mqt/bench/benchmarks/ghz_dynamic.py
burgholzer: I would also prefer to treat the num_qubits==1 case explicitly instead of implicitly adding a zero-bit classical register.

src/mqt/bench/benchmarks/ghz_dynamic.py
TomasVF: At first i had the reset operation as you mention it, but when running the test there is an error during transpilation for the following device:

`benchmark_name = 'ghz_dynamic'
device_name = 'ionq_aria_25'
gateset_name = 'rigetti'
input_value = 3`

Checking, the error comes from the reset operation.

src/mqt/bench/benchmarks/ghz_dynamic.py
TomasVF: That loop applies to all even qubits untill the end using the previous intermediate measurement, the first one is excluded as there is no measurement to be used there. Maybe i should clarify what it does in more detail in a comment.

burgholzer:
> > Dear @TomasVF, may I ask for the current status of this PR?
> 
> There were some corrections i had to do but i think everything is fixed now. I'm waiting for @burgholzer to check some of my answers as I was not too sure about some things.

I am probably missing something, but I don't see any answers from you. Could it be that these are still unpublished on your side?

TomasVF:
> > > Dear @TomasVF, may I ask for the current status of this PR?
> > 
> > 
> > There were some corrections i had to do but i think everything is fixed now. I'm waiting for @burgholzer to check some of my answers as I was not too sure about some things.
> 
> I am probably missing something, but I don't see any answers from you. Could it be that these are still unpublished on your side?


My mistake, I completly forgot to publish the review. It should already be visible now.

src/mqt/bench/benchmarks/ghz_dynamic.py
burgholzer: 👍🏼 no that some time has passed, I read the code again, and it makes much more sense now. I refactored it slightly to make the intent even clearer.

src/mqt/bench/benchmarks/ghz_dynamic.py
burgholzer: Yeah. I thought about this again and this solution might even be cleaner than the reset one, as the qubit is already measured anyway. Let's leave this as-is for now.

PR #916, opened 2026-06-03
Add IQPE benchmark

### LINKED ISSUE #797: ✨ Iterative Quantum Phase Estimation (IQPE)
### Iterative Quantum Phase Estimation (IQPE)

Iterative Quantum Phase Estimation (IQPE) is a variant of the standard Quantum Phase Estimation algorithm that replaces the quantum Fourier transform with repeated mid-circuit measuremets and classical feed-forward. Hence, it reduces the required number of qubits.

Instead of using multiple ancilla qubits, IQPE reuses a single ancilla qubit for several iterations. After each iteration, the ancilla is measured, and this outcome is used to classically control subsequent rotation gates.

**Relevant literature**
https://journals.aps.org/pra/pdf/10.1103/PhysRevA.76.030306
First proposed: https://arxiv.org/abs/quant-ph/9511026

### Proposed Solution

Because of its adaptive structure and practical relevance for near-term and fault-tolerant quantum computing, I propose to implement Iterative Quantum Phase Estimation as a new dynamic benchmark type in MQT Bench.

kpassito:
## Description

Adds an `iqpe` benchmark for Iterative Quantum Phase Estimation.

The benchmark uses one reusable measurement qubit and one target qubit, so `num_qubits` is currently fixed at 2. `num_bits` controls the estimation accuracy. Compared with the existing QPE variants, IQPE trades the wider phase register for additional dynamic circuit depth.

Fixes #797

## Tests

- `.venv/bin/pytest -o addopts='' tests/test_bench.py -k iqpe`: 8 passed
- `uvx nox -s lint`: passed
- `git diff --check`: passed

## AI disclosure

Codex assisted with implementation and local testing. I reviewed the changes and take responsibility for the submitted code.

The AI-assisted commits include the required footer:

`Assisted-by: GPT-5 Codex via Codex`

## Checklist

- [ ] The changes are fully tested and pass the CI checks.

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description.
- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.


docs/benchmark_selection.md
```diff
@@ -52,3 +52,10 @@ HTML(html)
 ```
 
 See the [benchmark description](https://www.cda.cit.tum.de/mqtbench/benchmark_description) for further details on the individual benchmarks.
+
+## Dynamic Iterative Phase Estimation
+
+The `iqpe` benchmark implements Iterative Quantum Phase Estimation as a dynamic circuit.
+Unlike the standard QPE benchmarks, it reuses one measurement ancilla and one target eigenstate qubit instead of allocating a full quantum phase register.
+Use `circuit_size=2`; the optional `num_bits` argument controls the number of mid-circuit measurement and classically controlled feedback iterations.
+The benchmark reinitializes the reusable ancilla with measurement-conditioned feedback so it remains compatible with targets that do not expose a native reset operation.
```
flowerthrower: ai blob - not in line with rest of the file

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,70 @@
...
+        num_qubits: Number of qubits in the returned circuit. Must be exactly 2:
+            one reusable measurement ancilla and one target eigenstate qubit.
+        num_bits: Number of phase-estimation iterations/classical bits.
+        phase: Target phase as a fraction of one full turn. Must be in ``[0, 1)``.
+
+    Returns:
+        QuantumCircuit: A dynamic IQPE circuit with mid-circuit measurements,
+            reset-by-feedback operations, and classically controlled rotations.
+    """
+    if num_qubits != 2:
```
flowerthrower, line 39: why constrain to 2 qubits only? should be possible for larger sizes too

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,70 @@
...
+        msg = "Number of qubits must be exactly 2 for IQPE."
+        raise ValueError(msg)
+    if num_bits < 1:
```
flowerthrower, line 42: this should not be the dynamic size of the benchmark (it should be num_qubits)

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- ✨ Add Iterative Quantum Phase Estimation (IQPE) benchmark.
```
flowerthrower: add pr number and author

flowerthrower (changes requested):
Dear @kpassito, thanks for the initial draft. Currently this looks like heavily AI generated. Please review our AI guidelines. I left a first quick review that you might want to address personally.

kpassito:
Thanks @flowerthrower, I went through the PR again and addressed the review points.

I removed the extra benchmark selection docs section, updated `iqpe` so `num_qubits` controls the circuit width, and fixed the changelog entry with the PR number and author.

I also reran the local checks listed in the PR description.

flowerthrower:
Dear @kpassito, thank you for your continued efforts. Please give us some time to review. We are currently internally discussing the best way forward for this Benchmark.

flowerthrower:
Dear @kpassito, thank you for your patience. We have now reached an internal agreement on the best approach to this benchmark. We have decided that we want to make this benchmark equivalent to the already implemented `qpe` variants after all. This does mean that `num_qubits` should remain a parameter limited to 2 qubits (for now) - sorry for the back-and-forth; ideally, that could have been avoided with a short discussion prior to implementation, as suggested in the original issue description. The result accuracy scales with `num_bits`, and the equivalence/trade-off between circuit width (normal `qpe`) and circuit depth (iterative version) should be highlighted in the benchmark description.

kpassito:
Thanks Patrick. I've updated the implementation to match the agreed direction: `num_qubits` is fixed at 2, `num_bits` controls accuracy, and the description now calls out the width/depth trade-off. I also resolved the two superseded review threads.

All lint and platform tests passed. The remaining CI failure is in Codecov's CLI signature verification (`No public key`), before the coverage upload, so it doesn't appear related to the PR changes. Could you please rerun that job when convenient and re-review?

burgholzer:
> All lint and platform tests passed. The remaining CI failure is in Codecov's CLI signature verification (`No public key`), before the coverage upload, so it doesn't appear related to the PR changes. Could you please rerun that job when convenient and re-review?

Just FYI. The codecov failure is not related to your changes. The codecov GitHub action broke just a few days ago. An update has been pushed already and a release of our CI workflows should come in today or tomorrow. Then, the workflows will be updated across all MQT repositories and the failures should be resolved.

flowerthrower:
LGMT; @burgholzer any final remarks?

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,69 @@
...
+    measurement = QuantumRegister(1, "measurement")
+    target = QuantumRegister(1, "target")
+    phase_bits = ClassicalRegister(num_bits, "phase")
+    qc = QuantumCircuit(measurement, target, phase_bits, name="iqpe")
+
+    qc.x(target[0])
+
+    for bit in reversed(range(num_bits)):
+        qc.h(measurement[0])
+        qc.cp(2 * math.pi * phase * (2**bit), measurement[0], target[0])
```
burgholzer, line 58: This expression will become astronomically large for even a moderate number of bits, and will likely break down around 64 bits (probably a bit earlier) due to integer bitwidth limitations.
Two ways to solve this:
- put an upper limit on the number of qubits/bits that can be used for this benchmark, or
- implement some clever modulo arithmetic that keeps this angle within `[-pi,pi)` (or equivalent)

I would personally prefer the second solution here as its cleaner and less surprising for the user.

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,69 @@
...
+
+        for correction_bit in range(bit + 1, num_bits):
+            with qc.if_test((phase_bits[correction_bit], 1)):
+                qc.rz(-math.pi / (2 ** (correction_bit - bit)), measurement[0])
```
burgholzer, line 62: The rotation angles here become exponentially small with increasing number of bits. It is probably worth introducing a cutoff at some point, where no more rotation gates are being generated if the rotation angle is too small. (Python will round to 0. at some point)

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,69 @@
...
+    trading circuit width for circuit depth.
+
+    Arguments:
+        num_qubits: Number of qubits of the returned quantum circuit. Must be exactly 2.
+        num_bits: Number of measured phase bits and dynamic iterations.
+        phase: Target phase as a fraction of one full turn. Must be in ``[0, 1)``.
+
+    Returns:
+        QuantumCircuit: The constructed IQPE circuit.
+    """
```
burgholzer, line 38: This is different than the corresponding QPE benchmark (https://github.com/munich-quantum-toolkit/bench/blob/main/src/mqt/bench/benchmarks/qpeexact.py and https://github.com/munich-quantum-toolkit/bench/blob/main/src/mqt/bench/benchmarks/qpeinexact.py), which is supposed to be the non-dynamic version of the benchmark here.

The semantics as well as the implementation here should mirror the QPE code, but make it iterative.
As such, I would argue that the benchmark should have the signature
```python
def create_circuit(num_qubits: int, exact: bool = True)
```
It should create quantum registers `q` and `psi`, as well as a classical register `c`. Then it should generate an appropriate phase for the benchmark that is either exactly representable or not depending on the input parameter and then perform the IQPE algorithm.

It may be worth to call the classical register `"meas"` to align with the convention that Qiskit uses when calling `qc.measure_all()`.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- ✨ Add Iterative Quantum Phase Estimation (IQPE) benchmark ([#916]) ([**@kpassito**])
```
burgholzer, line 14: In order for the links to work, you need to add the PR link and your Author page link to the bottom of the changelog.

burgholzer (changes requested):
Hey @kpassito 👋🏼 

Thanks for your continued work on this. Highly appreciated.
This is not yet quite where we want it to be so that it can be merged. You'll find more detailed comments inline in the respective file. I hope these make sense.

The overall goal should be to align the implementation here with the QPE implementation(s) so that this benchmark "simply" becomes the dynamic circuit variant of the QPE algorithm.

Don't hesitate to ask questions in case they pop up!

burgholzer:
Thanks @kpassito for your interest in contributing to MQT Bench. In the meantime, another solution has been created and merged. As such, I am afraid there is not much to do for us here than to close the PR.
I hope you'll consider contributing to other issues in the future.

PR #925, opened 2026-06-11
✨ Add Iterative Quantum Phase Estimation (IQPE)

### LINKED ISSUE #797: ✨ Iterative Quantum Phase Estimation (IQPE)

JohannesWittmann9:
## Description

Implementation of the Iterative Quantum Phase Estimation (IQPE) benchmark. The code is inspired by existing benchmarks `qpeexact` and `qpeinexact`, combining the two with an additional parameter `exact: bool`.
The circuit accepts a minimum of `num_qubits=2`, which is split into 1 ancillar qubit and I calculate `num_iterations = num_qubits - 1` iterations. The split of the `num_qubits` parameter was chosen to be consistent with the input and output of `qpeexact` and `qpeinexact`. 
It extracts the phase over `num_iterations` rounds using mid-circuit measurement with classical feedforward (if_test). In the inexact case an extra bit is added so the available iterations cannot represent the phase exactly. 

A few important details:
- The implementation is based on information found in this [Dobšíček et al.](https://journals.aps.org/pra/pdf/10.1103/PhysRevA.76.030306)
- To keep calculations from exploding, our phase is reduced to a fraction in [0, 1) and the angle is kept within (-pi, pi]
- For rotation angles, if they are close to 0, rotations are skipped.

Fixes #797

## Checklist

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/bench/blob/main/docs/ai_usage.md).
- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

### DIFF
--- a/src/mqt/bench/benchmarks/iqpe.py
+++ b/src/mqt/bench/benchmarks/iqpe.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""IQPE benchmark definition."""
+
+from __future__ import annotations
+
+import random
+from fractions import Fraction
+
+import numpy as np
+from qiskit.circuit import ClassicalRegister, QuantumCircuit, QuantumRegister
+
+from ._registry import register_benchmark
+
+
+@register_benchmark("iqpe", description="Iterative Quantum Phase Estimation (IQPE)")
+def create_circuit(num_qubits: int, exact: bool = True) -> QuantumCircuit:
+    """Returns a quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
+
+    Arguments:
+        num_qubits: Number of qubits of the returned quantum circuit. Must be at least 2.
+        exact: Whether to use the exact version of the algorithm.
+
+    Returns:
+        QuantumCircuit: A quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
+    """
+    if num_qubits <= 1:
+        msg = "Number of qubits must be at least 2 for IQPE."
+        raise ValueError(msg)
+
+    num_iterations = num_qubits - 1  # one ancilla qubit, rest for iterations
+    ancilla = QuantumRegister(1, "q")
+    psi = QuantumRegister(1, "psi")
+    c = ClassicalRegister(num_iterations, "c")
+    qc = QuantumCircuit(ancilla, psi, c, name="iqpe")
+
+    # get random n-bit string as target phase
+    random.seed(10)
+    theta = 0
+    if exact:
+        while theta == 0:
+            theta = random.getrandbits(num_iterations)
+        lam = Fraction(theta, 1 << num_iterations)  # phase to be estimated
+
+    elif not exact:
+        while theta == 0 or (theta & 1) == 0:  # Ensure an odd theta so phase cannot be exactly represented with n bits
+            theta = random.getrandbits(num_iterations)
+        lam = Fraction(theta, 1 << (num_iterations + 1))  # phase to be estimated, with extra bit
+
+    qc.x(psi)  # prepare |1> state
+
+    for i in range(num_iterations, 0, -1):  # start from n
+        qc.reset(ancilla[0])  # reset ancilla qubit
+        qc.h(ancilla[0])  # put ancilla in superposition
+
+        frac = (lam * (1 << (i - 1))) % 1  # exact Fraction in [0, 1)
+        angle = 2 * np.pi * float(frac)  # in [0, 2*pi)
+        if angle > np.pi:
+            angle -= 2 * np.pi  # shift into (-pi, pi]
+        qc.cp(angle, psi, ancilla[0])
+
+        for meas_idx in range(i, num_iterations):  # bits already measured this run
+            with qc.if_test((c[meas_idx], 1)):
+                rotation_angle = -2 * np.pi / (1 << (meas_idx - (i - 1) + 1))
+                if abs(rotation_angle) > 1e-10:  # avoid adding negligible gates
+                    qc.rz(rotation_angle, ancilla[0])
+
+        qc.h(ancilla[0])
+        qc.measure(ancilla[0], c[i - 1])
+
+    return qc
--- a/src/mqt/bench/benchmarks/iqpe.py
+++ b/src/mqt/bench/benchmarks/iqpe.py
@@ -20,12 +20,16 @@ from ._registry import register_benchmark
 
 
 @register_benchmark("iqpe", description="Iterative Quantum Phase Estimation (IQPE)")
-def create_circuit(num_qubits: int, exact: bool = True) -> QuantumCircuit:
+def create_circuit(
+    num_qubits: int, exact: bool = True, rotation_threshold: float = 1e-15, seed: int = 10
+) -> QuantumCircuit:
     """Returns a quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
 
     Arguments:
         num_qubits: Number of qubits of the returned quantum circuit. Must be at least 2.
         exact: Whether to use the exact version of the algorithm.
+        rotation_threshold: Threshold for adding rotation gates in the feedback step. Rotation gates with smaller angles will be omitted to reduce circuit complexity.
+        seed: Seed for the random number generator.
 
     Returns:
         QuantumCircuit: A quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
@@ -41,8 +45,8 @@ def create_circuit(num_qubits: int, exact: bool = True) -> QuantumCircuit:
     qc = QuantumCircuit(ancilla, psi, c, name="iqpe")
 
     # get random n-bit string as target phase
-    random.seed(10)
     theta = 0
+    random.seed(seed)
     if exact:
         while theta == 0:
             theta = random.getrandbits(num_iterations)
@@ -67,7 +71,7 @@ def create_circuit(num_qubits: int, exact: bool = True) -> QuantumCircuit:
         for meas_idx in range(i, num_iterations):  # bits already measured this run
             with qc.if_test((c[meas_idx], 1)):
                 rotation_angle = -2 * np.pi / (1 << (meas_idx - (i - 1) + 1))
-                if abs(rotation_angle) > 1e-10:  # avoid adding negligible gates
+                if abs(rotation_angle) > rotation_threshold:  # avoid adding negligible gates
                     qc.rz(rotation_angle, ancilla[0])
 
         qc.h(ancilla[0])
--- a/src/mqt/bench/benchmarks/iqpe.py
+++ b/src/mqt/bench/benchmarks/iqpe.py
@@ -80,6 +80,7 @@ def create_circuit(
 
         qc.h(ancilla[0])
         qc.measure(ancilla[0], c[i - 1])
-        qc.reset(ancilla[0])
+        with qc.if_test((c[i - 1], 1)):
+            qc.x(ancilla[0])  # reset ancilla for next iteration
 
     return qc
--- a/src/mqt/bench/benchmarks/iqpe.py
+++ b/src/mqt/bench/benchmarks/iqpe.py
@@ -56,7 +56,6 @@ def create_circuit(num_qubits: int, exact: bool = True) -> QuantumCircuit:
     qc.x(psi)  # prepare |1> state
 
     for i in range(num_iterations, 0, -1):  # start from n
-        qc.reset(ancilla[0])  # reset ancilla qubit
         qc.h(ancilla[0])  # put ancilla in superposition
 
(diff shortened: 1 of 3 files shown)


src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,78 @@
...
+        raise ValueError(msg)
+
+    num_iterations = num_qubits - 1  # one ancilla qubit, rest for iterations
+    ancilla = QuantumRegister(1, "q")
+    psi = QuantumRegister(1, "psi")
+    c = ClassicalRegister(num_iterations, "c")
+    qc = QuantumCircuit(ancilla, psi, c, name="iqpe")
+
+    # get random n-bit string as target phase
+    random.seed(10)
```
burgholzer: It could make sense to make the seed configurable so one could at least in principle generate multiple instances.
Can simply be added as an additional function parameter defaulting to `10`.

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,78 @@
...
+        while theta == 0 or (theta & 1) == 0:  # Ensure an odd theta so phase cannot be exactly represented with n bits
+            theta = random.getrandbits(num_iterations)
+        lam = Fraction(theta, 1 << (num_iterations + 1))  # phase to be estimated, with extra bit
+
+    qc.x(psi)  # prepare |1> state
+
+    for i in range(num_iterations, 0, -1):  # start from n
+        qc.h(ancilla[0])  # put ancilla in superposition
+
+        frac = (lam * (1 << (i - 1))) % 1  # exact Fraction in [0, 1)
```
burgholzer, line 70: This still computes `1 << (i-1)`, right?
So it is still somewhat limited in scalability, right?

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,78 @@
...
+        angle = 2 * np.pi * float(frac)  # in [0, 2*pi)
+        if angle > np.pi:
+            angle -= 2 * np.pi  # shift into (-pi, pi]
+        qc.cp(angle, psi, ancilla[0])
+
+        for meas_idx in range(i, num_iterations):  # bits already measured this run
+            with qc.if_test((c[meas_idx], 1)):
+                rotation_angle = -2 * np.pi / (1 << (meas_idx - (i - 1) + 1))
```
burgholzer: This also still computes `(1 << (meas_idx - (i - 1) + 1))` which may run into overflow problems, right?

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,78 @@
...
+                if abs(rotation_angle) > 1e-10:  # avoid adding negligible gates
```
burgholzer: Would it make sense to make this threshold configurable? Could be another function parameter defaulting to `1e-10`.
Maybe the default should also be a little lower (1e-15 for example).

burgholzer (changes requested):
Thanks @JohannesWittmann9 🙏🏼 This looks very clean already. I had a look through the changes and have a couple of nitpick questions/comments that you will find inline.
None of them should be particularly hard to address 😌 
Pretty sure that this can be merged after the comments are addressed.

src/mqt/bench/benchmarks/iqpe.py
JohannesWittmann9: Yes, that is a good idea, I also thought of that. I just fixed it, because it is also like that in qpe benchmarks. Maybe an idea to change it there as well? 😄

burgholzer: Yeah. But let's reserve that for a follow-up PR to keep this one self-contained.

src/mqt/bench/benchmarks/iqpe.py
JohannesWittmann9: Yes, while the `1e-10` is already really small (I couldn't reach the threshold till a high amount of iterations), adding it as a configurable param would be a good idea and needed for scalability and transparency I guess.

src/mqt/bench/benchmarks/iqpe.py
JohannesWittmann9: Yes, this indeed computes `1 << (i-1)`. However I don't think this is a scalability issue, as the calculation itself is not the problem (Please correct me if wrong :)).
I think the problem would be not scaling it down before applying the controlled phase gate.

burgholzer: If I am not mistaken, `1 << (i-1)` is an integer computation. As such, assuming 64bit integers, it will overflow for `i>=65` and yield wrong results.

JohannesWittmann9: Python integers are arbitrary precision, so this shouldn't be a problem.
The multiplication with lam should also be good, because it is a Fraction so no overflows or precision problems.
Correct me if I'm wrong or thumbs up if correct. :D

src/mqt/bench/benchmarks/iqpe.py
JohannesWittmann9: The problem I would see is that the denominator `(1 << (meas_idx - (i - 1) + 1))` can get really big for high `num_iterations`. With the division, we transform it into a float, which could lead to overflow problems for large number of iterations. The angle itself gets closer to 0.
Maybe we could limit the amount of phase correction steps, maybe also bound to the minimal rotation angle?

src/mqt/bench/benchmarks/iqpe.py
burgholzer: Just checked, and you are absolutely right. Learned something new today!
This comment can be resolved 👍🏻

src/mqt/bench/benchmarks/iqpe.py
burgholzer: Yeah, probably there should be some
Kind of guard related to the approximation precision.

JohannesWittmann9: Maybe something like this: For an angle smaller than our threshold `2 * pi / 2^k < rotation_threshold` we can calculate `k > log2(2 * pi / rotation_threshold)` so we compute the `k` based on the `rotation_threshold` param.
Then our `(meas_idx - (i - 1) + 1))` needs to be smaller or equal `k`.
So if `k = 50` we only apply 50 rotations max.
Let me know what you think :)

Alternative: I think we get the overflow for a very large k, we could also set an upper bound. However might not be as transparent.

(Edit: srry had a small error in the formular)

burgholzer: Maybe we are thinking about this too complicated. How about we simply apply the same `if(angle >= tolerance) apply rotation`?

JohannesWittmann9: Yes, you are right, I was overengineering. 
I wrote a little test script to see when the `-2 * np.pi / (1 << (meas_idx - (i - 1) + 1))` produces an `OverflowError: int too large to convert to float`. Just fyi, happens at `(meas_idx - (i - 1) + 1) >= 1024`.

However, there is a simple fix, which is again a feature of python. If we do it like this `-2 * np.pi * (2.0 ** -exponent)`, the float `(2.0 ** -exponent)` becomes really small for large exponents, but eventually if we do not have any bits left will just be rounded to `0.0`. 

(Edit: Alternative we can also write it `-2 * np.pi * Fraction(1, 1 << exponent)`, maybe that is cleaner with respect to the implementation)

After we then have our if statement with the rotation_threshold as it is now.

I think that should clear it. Feel free to give a thumbs up, then I will implement it and commit the final fix. :)

burgholzer: `-2 * np.pi * Fraction(1, 1 << exponent)`
feels cleanest 👍🏼 
can be used in both places where a rotation angle is currently being computed.

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,83 @@
...
+
+@register_benchmark("iqpe", description="Iterative Quantum Phase Estimation (IQPE)")
+def create_circuit(
+    num_qubits: int, exact: bool = True, rotation_threshold: float = 1e-15, seed: int = 10
+) -> QuantumCircuit:
+    """Returns a quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
+
+    Arguments:
+        num_qubits: Number of qubits of the returned quantum circuit. Must be at least 2.
+        exact: Whether to use the exact version of the algorithm.
```
burgholzer:
```suggestion
        num_qubits: Controls the precision of the phase estimation algorithm. Must be at least 2.
        exact: Whether to use a target phase that is exactly representable with the chosen precision.
```

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,83 @@
...
+from ._registry import register_benchmark
+
+
+@register_benchmark("iqpe", description="Iterative Quantum Phase Estimation (IQPE)")
+def create_circuit(
+    num_qubits: int, exact: bool = True, rotation_threshold: float = 1e-15, seed: int = 10
+) -> QuantumCircuit:
+    """Returns a quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.
+
+    Arguments:
```
burgholzer, line 33:
```suggestion
    """Returns a quantum circuit implementing the Iterative Quantum Phase Estimation algorithm.

    In contrast to the regular QPE algorithm, the @p num_qubits argument does not influence the final number of qubits of the circuit.
    The circuit will always use exactly two qubits, but `num_qubits-1` iterations of the phase estimation loop.
    This way, transforming the IQPE benchmark circuit to a straight-line circuit (by replacing reset operations with new qubits and applying the deferred measurement principle) results in the equivalent QPE benchmark circuit configured with the same value for `num_qubits`.
    The name `num_qubits` is kept for consistency with the other benchmark generation functions.

    Arguments:
```

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,83 @@
...
+
+    qc.x(psi)  # prepare |1> state
+
+    for i in range(num_iterations, 0, -1):  # start from n
+        qc.h(ancilla[0])  # put ancilla in superposition
+
+        frac = (lam * (1 << (i - 1))) % 1  # exact Fraction in [0, 1)
+        angle = 2 * np.pi * frac  # in [0, 2*pi)
+        if angle > np.pi:
+            angle -= 2 * np.pi  # shift into (-pi, pi]
```
burgholzer:
```suggestion
        angle = 2 * np.pi * frac  # in [0, 2*pi)
```
Should not be necessary.

src/mqt/bench/benchmarks/iqpe.py
```diff
@@ -0,0 +1,83 @@
...
+            with qc.if_test((c[meas_idx], 1)):
+                exponent = meas_idx - (i - 1) + 1
+                rotation_angle = -2 * np.pi * Fraction(1, 1 << exponent)
+                if abs(rotation_angle) > rotation_threshold:  # avoid adding negligible gates
+                    qc.rz(rotation_angle, ancilla[0])
+
+        qc.h(ancilla[0])
+        qc.measure(ancilla[0], c[i - 1])
+        with qc.if_test((c[i - 1], 1)):
+            qc.x(ancilla[0])  # reset ancilla for next iteration
```
burgholzer:
```suggestion
        qc.reset(ancilla[0])
```
Some platforms implement more efficient reset techniques. This should be the "pure" algorithm level implementation.

burgholzer (approved):
Nice work 🙌🏼 
I just have a couple of minor tweaks to the wording and the implementation in general, but I'll simply apply these in a batch after the review. Then this should be good to go 🚀

burgholzer:
~~Ah, the `reset` operation is not declared as supported by some of the gate sets.
Let me quickly fix that. There is no reason not to expose it.~~

Nevermind. That's too much of a change for too little gain. The if_test solution is just fine here.

burgholzer:
Congrats @JohannesWittmann9 on the first contribution to the MQT and a successfully closed UnitaryHACK issue 🙌

JohannesWittmann9:
Great, thank you! @burgholzer 
Was really pleasant working with you. 😄 
Already looking for the next bounty. 👁️

PR #935, opened 2026-06-18
🔧 Maintenance round

denialhaag:
## Description

This PR makes several maintenance updates:

- Fix `required-checks-pass` job
- Use `client-id` instead of `app-id`
- Improve support for `CLAUDE.md`
- Configure Read the Docs to use Python 3.14 for colorful docs
- Configure Read the Docs to natively use `uv`

## Checklist

- [x] ~~I have added appropriate tests that cover the new/changed functionality.~~
- [x] ~~I have updated the documentation to reflect these changes.~~
- [x] ~~I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.~~
- [x] ~~I have added migration instructions to the upgrade guide (if needed).~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,11 +88,11 @@ jobs:
         with:
           allowed-skips: >-
             ${{
-              fromJSON(needs.change-detection.outputs.run-python-tests)
-              && '' || 'python-tests,python-coverage,python-linter,'
+              !fromJSON(needs.change-detection.outputs.run-python-tests)
+              && 'python-tests,python-coverage,python-linter,' || ''
             }}
             ${{
-              fromJSON(needs.change-detection.outputs.run-cd)
-              && '' || 'build-sdist,build-wheel'
+              !fromJSON(needs.change-detection.outputs.run-cd)
+              && 'build-sdist,build-wheel' || ''
             }}
           jobs: ${{ toJSON(needs) }}
--- a/.github/workflows/templating.yml
+++ b/.github/workflows/templating.yml
@@ -19,7 +19,7 @@ jobs:
       - id: create-token
         uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         with:
-          app-id: ${{ secrets.APP_ID }}
+          client-id: ${{ secrets.APP_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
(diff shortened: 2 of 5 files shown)


denialhaag:
I'm very confused why `check-sdist` is suddenly failing. I retriggered the last job on `main`, and it's failing on `main` as well. MQT Predictor is also affected, but MQT DDSIM and MQT QCEC are fine. Any ideas, @burgholzer?

burgholzer:
Could be related to the build backend. The python projects are using the uv one. Either an update there, or an update with check-sdist, or both.
Worth checking the changelogs.

Could potentially be fixed by adding the offending files to the gitignore file.

denialhaag:
> Could be related to the build backend. The python projects are using the uv one. Either an update there, or an update with check-sdist, or both. Worth checking the changelogs.

I'm assuming it's https://github.com/henryiii/check-sdist/pull/161, which was released yesterday. 🤔 

> Could potentially be fixed by adding the offending files to the gitignore file.

Wouldn't that defeat the purpose of `--inject-junk`? 🤔

denialhaag:
> > Could potentially be fixed by adding the offending files to the gitignore file.
> 
> Wouldn't that defeat the purpose of `--inject-junk`? 🤔

Okay, my brain seems to be a bit slow today... The purpose of `--inject-junk` is to point us to adding the files to `.gitignore`... 🙃

burgholzer:
> > Could be related to the build backend. The python projects are using the uv one. Either an update there, or an update with check-sdist, or both. Worth checking the changelogs.
> 
> 
> 
> I'm assuming it's https://github.com/henryiii/check-sdist/pull/161, which was released yesterday. 🤔 
> 
> 
> 
> > Could potentially be fixed by adding the offending files to the gitignore file.
> 
> 
> 
> Wouldn't that defeat the purpose of `--inject-junk`? 🤔 

Ah yeah. That would make sense.

Actually no. These junk files should never be commited to version control. The default option to achieve that would be to add them to the ignore file.

denialhaag:
@burgholzer, thanks for the help! This should be ready now.

We should consider adding `.gitignore` to the templates. Generally speaking, a project-type-specific template (i.e., Python or C++) should be sufficient. 🤔

burgholzer (approved):
LGTM 👍🏻

I am all for adding the ignore file to the templates. We can just use the superset of the MQT repos. I don't even think there needs to be a distinction between c++ and python.
The _version.py file may need some special treatment (or a better regex).

Issue #924, opened 2026-06-09
✨ Add expected noise-free outputs for circuits


grageragarces:
### Problem Statement

It would be really useful if MQT Bench circuits came with their expected noise-free results (an ideal state vector or output probability distribution) so that downstream users can validate a experiments against an expected truth.
For small circuits or things like GHZ and W it doesn't matter, but for the more involved algorithms expected results aren't obvious from the circuit alone. For instance, in Grover,  without knowing how the generator constructs the oracle, there's no way to tell which state the circuit is meant to amplify, so there's nothing to test against.

### Proposed Solution

If you have them add them in some format. Maybe json would be good?
`grover_indep_5.qasm` → `grover_indep_5_reference.json`

burgholzer:
Thanks for the suggestion @grageragarces 🙏🏼 
This is something that has come up before in discussions and that would generally be helpful.
For many of the benchmarks, it is moderately easy to write a function that verifies a distribution of measurement counts against the expected distribution, or verifies that the "right" solution has been chosen most often.
There are two things that make this tricky:
- most of the benchmarks in MQT Bench are scalable in the sense that they can be instantiated for arbitrarily many qubits. This rules any kind of solution out that describes the reference as a dense state vector.
- only some of the benchmarks have a sparse output distribution with a clear thing to look for. Some are equal superpositions of a bunch of states; some are almost random. It is fairly hard to find general "acceptance criteria".

Would you have any ideas in mind for how the expected output could be structured?

grageragarces:
I think to calculate the basic metrics (total variation distance (TVD), success probability, Hellinger fidelity, ...) we can just build a pointwise evaluator `P(x)` that we only need to query on the finite set of bitstrings we measure. Since the measured distribution `Q` should be finite, we should be good even at scale.

For fidelity this works because terms with `Q(x) = 0` vanish, so the sum collapses onto `supp(Q)`:

```
F(P,Q) = (Σ_x √(P(x)·Q(x)))²  =  (Σ_{x ∈ supp(Q)} √(P(x)·Q(x)))²
```

And TVD comes out too as long as `P` is normalized:

```
TVD(P,Q) = ½ Σ_x |P(x) − Q(x)|
         = ½ [ Σ_{x∈O} |P(x) − Q(x)|  +  Σ_{x∉O} P(x) ]
         = ½ [ Σ_{x∈O} |P(x) − Q(x)|  +  (1 − Σ_{x∈O} P(x)) ]
```

[...]

PR #941, opened 2026-06-26
Add reference spec framework for noise-free benchmark outputs (#924)

grageragarces:
## Summary

Implements the reference specification framework proposed in #924.
Each benchmark can now expose a `create_reference` function that returns
a compact, JSON-serialisable description of the ideal output distribution —
no dense state vectors required.

### Distribution kinds

| Kind | Example | What is stored |
|---|---|---|
| `sparse` | GHZ, BV, Grover | `{bitstring: prob}` table |
| `uniform` | W state | predicate + support size (P(x) = 1/size if predicate holds) |
| `simulate` | Random circuit | advisory `max_qubits` — caller runs statevector sim |
| `none` | QAOA, VQE | reason string — no compact form exists |

An optional `objective` field carries the semantic answer (hidden string,
marked states, phase, …) separately from the distribution.

### Implemented benchmarks

`ghz`, `wstate`, `bv`, `dj`, `grover`, `qpeexact`, `randomcircuit`

### Public API

```python
from mqt.bench.benchmarks import get_reference_spec, has_reference

spec = get_reference_spec("grover", 5)
print(spec.to_dict())   # JSON-ready dict


grageragarces:
Hey @burgholzer! 👋

I've opened a draft PR implementing the reference spec framework proposed in #924. 
Here's what I have added:

- A new _reference.py module with the type hierarchy we sketched: SparseReference, UniformReference, SimulateReference, NoneReference, ObjectiveSpec, and a top-level ReferenceSpec with a to_dict() method for JSON serialisation.
- A register_reference decorator that mirrors register_benchmark — each benchmark file can optionally define a companion create_reference(num_qubits, ...) function. 
- Public API: get_reference_spec("ghz", 5) and has_reference("grover"). 
- Reference specs for 7 benchmarks as worked examples of each kind: ghz (sparse), wstate (uniform), bv, dj, qpeexact (all sparse/deterministic), grover (sparse + marked-states objective), randomcircuit (simulate).

The remaining benchmarks need someone who knows the generator internals to fill in the right answer. Some thoughts on the obvious ones:

- QFT / QFT-entangled: is the reference the QFT of a known input state, or do these go under simulate?
- AE / QPE-inexact: inexact phase estimation has a distribution over neighbouring bitstrings; happy to implement if you can confirm the target phase used.
- Adder circuits: these are deterministic given fixed inputs; what operand values does the generator use?
- QAOA / VQE: I've left these as NoneReference since the output depends on optimised parameters. Is that the right call, or would you prefer a SimulateReference so at least small instances can be validated?
- Graph state / random circuits beyond ~30q: SimulateReference with an advisory max_qubits field seems right. Alt we could just have a NoneReference.

Any guidance on the above would let me fill in the rest. 


Also very happy to adjust the schema if something doesn't work quite right!
Perhaps having the references in a subfolder might make it cleaner?

grageragarces:
will give it a day for the tests to fail before trying to fix them :)

